PR #3, opened 2015-02-19
A 2bit parser with tests using Flow

danvk:
See README.md for details on how to work with this.

One notable difference from the organization of CycleDash is that tests are always run in a browser (`grunt test` uses PhantomJS). I think this is the Right Way to do things for a few reasons:

1. Our JS ultimately runs in a browser, so it makes sense to test it in a browser.
2. We don't have to effectively re-build a browser in our tests by shimming things like the DOM and XMLHttpRequest in node.js. This also means that libraries like jQuery, sinon and D3 will behave more predictably.
3. It will let us do true cross-browser testing, e.g. with [saucelabs](https://saucelabs.com/u/algolia-faux-jax)


src/readableview.js
```diff
@@ -0,0 +1,67 @@
...
+    return result;
+  }
+
+  // Returns the number of bytes remaining in the buffer.
+  bytesRemaining(): number {
+    return this.dataView.byteLength - this.offset;
+  }
+
+  // Returns the current offset in the buffer.
+  tell(): number {
```
ihodes: Maybe `getCurrentOffset()` or `offset()` or `getOffset()`?

src/remotefile.js
```diff
@@ -0,0 +1,73 @@
...
+  getFromNetwork(start: number, stop: number): Q.Promise<DataView> {
+    var deferred = Q.defer();
+
+    var xhr = new XMLHttpRequest();
+    xhr.open('GET', this.url);
+    xhr.responseType = 'arraybuffer';
+    xhr.setRequestHeader('Range', `bytes=${start}-${stop}`);
+    var remoteFile = this;
+    xhr.onload = function(e) {
+      console.log('response came back');
```
ihodes: nix console.log?

src/readableview.js
danvk: `tell` is a pretty standard name for a method which gives you your current position in a file, e.g. http://www.tutorialspoint.com/python/file_tell.htm

src/remotefile.js
danvk: Done

src/readableview.js
ihodes: Yeah, and in perl and others. I believe Java uses `position()` for buffers, on the other hand. I'm fine with `tell`, just a suggestion.

src/2bit.js
```diff
@@ -0,0 +1,204 @@
...
+  // Remove base pairs from the end if the sequence terminated mid-byte.
+  basePairs.length = numBasePairs;
+  return basePairs;
+}
+
+/**
+ * Change base pairs to 'N' where the SequenceRecord dictates.
+ * This modifies the basePairs array in-place.
+ */
+function markUnknownDNA(basePairs: string[], dnaStartIndex: number, sequence: SequenceRecord): Array<string> {
```
ihodes: `string[]` = `Array<string>`?

src/readableview.js
```diff
@@ -0,0 +1,67 @@
+/** @flow */
```
ihodes: rename to ReadableView.js or readable_view?.js

src/remotefile.js
```diff
@@ -0,0 +1,72 @@
+/** @flow */
```
ihodes: maybe rename (like readable view)?

src/2bit.js
```diff
@@ -0,0 +1,203 @@
...
+      // The masks can be quite large (~2MB for chr1) and we mostly don't care
+      // about them.  So we ignore them, but we do need to know their length.
+      maskBlockCount = bytes.readUint32();
+      // maskBlockCount maskBlockStarts = bytes.readUint32Array(maskBlockCount),
+      // maskBlockSizes = bytes.readUint32Array(maskBlockCount),
+      // reserved = bytes.readUint32();
+
+  var offset = bytes.tell() + 8 * maskBlockCount + 4;
+
+  // DNA information comes after this.
```
ihodes: Not sure what this comment means?

src/2bit.js
ihodes: As someone who doesn't know the 2bit standard… So this is taking the output of unpackDNA and changing some random letter to an 'N' where the SequenceRecord says they should be?

test/fake-xhr.js
```diff
@@ -0,0 +1,92 @@
+/** @flow */
```
ihodes: similar naming thing; is a file just exports a class, maybe the file should be named like the class e.g. FakeXHR (similar to how we export components in CycleDash)?

ihodes:
This looks great. I need to read the 2bit spec to fully review this, and I should anyway, but this looks good.

src/readableview.js
danvk: The [Google JS Style Guide](https://google-styleguide.googlecode.com/svn/trunk/javascriptguide.xml#Naming__body) says:

> In general, use ... filenameslikethis.js.
> Filenames should be all lowercase in order to avoid confusion on case-sensitive platforms. Filenames should end in .js, and should contain no punctuation except for - or _ (prefer - to _).

So it should be either `readableview.js` or `readable-view.js`. Any preference?

src/2bit.js
danvk: Removed

src/readableview.js
ihodes: I know the GSG is against it, but we *are* currently using CamelCase in file names in CycleDash, and it makes even more sense, I'd think, when using ES6-style classes like we're doing here and exporting only them from the module. I think GSG is a bit out of date when it comes to e.g. flow, ES6, and React.js. If you still to use all lower-case, then my preference is with `readable-view.js`.

danvk:
Comments all addressed. I added a link to the [2bit spec](http://genome.ucsc.edu/FAQ/FAQformat.html#format7) to `2bit.js`.

Let's settle on a file naming convention and then we should be good to go.

src/2bit.js
danvk: Correct on both counts.

(I'm surprised `string[]` compiled -- I'm unclear if that's a valid syntax in Flow.)

src/readableview.js
danvk: It looks like Facebook uses Java-style `CamelCase.js` in all their React apps, so I'm OK doing the same. I didn't see an official Facebook style guide.

GSG definitely doesn't pre-date the idea of classes in JS or files that export a single class. The rationale for `lowercase.js` is that if you're developing on a case-insensitive file system (like a mac), then your site may work locally but not when you deploy (on linux). I don't see that as particularly relevant for us, since we're using an explicit build step.

PR #6, opened 2015-02-24
Display the reference genome

danvk:
Here it is in action:

screen shot 2015-02-23 at 7 00 09 pm

Nothing too fancy, but it works!

There were a few Flow warts that I ran into:

  - Flow can't track types through `this.props`—the `React.PropTypes` type system isn't precise enough.
  - There's an [`import type`][1] statement coming very soon, but it won't import type aliases. "Soon", they say. For now, I'm putting shared type aliases in a global file.

This version of the code doesn't have much to say about how data will be organized eventually—thoughts on that would be much appreciated!

[1]: http://flowtype.org/blog/2015/02/18/Import-Types.html

### DIFF
--- a/src/GenomeTrack.js
+++ b/src/GenomeTrack.js
@@ -1,0 +1,38 @@
+/**
+ * A track which displays a reference genome.
+ * @flow
+ */
+
+var React = require('react'),
+    types = require('./types');
+
+var GenomeTrack = React.createClass({
+  propTypes: {
+    range: types.GenomeRange,
+    basePairs: React.PropTypes.string
+  },
+  render: function(): any {
+    if (!this.props.range) {
+      return <EmptyTrack />;
+    }
+    var range = this.props.range;
+    var rangeLength = range.limit - range.start;
+    if (rangeLength > 200) {
+      return <EmptyTrack />;
+    }
+
+    if (!this.props.basePairs) {
+      return <div>no data</div>;
+    }
+
+    return <div>{this.props.basePairs}</div>;
+  }
+});
+
+var EmptyTrack = React.createClass({
+  render: function() {
+    return <div>Zoom in to see bases</div>
+  }
+});
+
+module.exports = GenomeTrack;
--- a/src/GenomeTrack.js
+++ b/src/GenomeTrack.js
@@ -12,10 +12,10 @@ var GenomeTrack = React.createClass({
     basePairs: React.PropTypes.string
   },
   render: function(): any {
-    if (!this.props.range) {
+    var range = this.props.range;
+    if (!range) {
       return <EmptyTrack />;
     }
-    var range = this.props.range;
     var rangeLength = range.limit - range.start;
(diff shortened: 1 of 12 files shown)


src/Root.js
```diff
@@ -0,0 +1,51 @@
...
+  },
+  getInitialState: function(): any {
+    return {
+      contigList: ['1', '2', '3'],
+      range: null,
+      basePairs: null
+    }
+  },
+  componentDidMount: function() {
+    // Note: flow is unable to infer this type through `this.propTypes`.
```
ihodes: As discussed, ugh.

src/Root.js
```diff
@@ -0,0 +1,51 @@
...
+    var ref: TwoBit = this.props.referenceSource;
+    ref.getContigList().then(contigList => {
+      this.setState({contigList});
+    });
+  },
+  handleRangeChange: function(newRange: GenomeRange) {
```
ihodes: Where is `GenomeRange` coming from? Should this be `types.GenomeRange`?

src/Root.js
```diff
@@ -0,0 +1,51 @@
...
+    // TODO: make this an "import type" when react-tools 0.13.0 is out.
+    TwoBit = require('./TwoBit');
+
+var Root = React.createClass({
+  propTypes: {
+    referenceSource: React.PropTypes.object.isRequired
+  },
+  getInitialState: function(): any {
+    return {
+      contigList: ['1', '2', '3'],
```
ihodes: Any reason this isn't just blank, as it'll be filled in by `componentDidMount`?  Flash of no-content type consideration?

src/GenomeTrack.js
```diff
@@ -0,0 +1,38 @@
...
+var GenomeTrack = React.createClass({
+  propTypes: {
+    range: types.GenomeRange,
+    basePairs: React.PropTypes.string
+  },
+  render: function(): any {
+    if (!this.props.range) {
+      return <EmptyTrack />;
+    }
+    var range = this.props.range;
```
ihodes: Move this above the previous `if` statement, and `s/this.props.range/range/`in that statement?

ihodes:
Looks good. Flow has some tricky warts, doesn't it?

src/Root.js
danvk: `types.GenomeRange` is the `React.PropTypes` version of this type. The Flow version of the type is in `types/types.js`. I was unable to namespace the types in that file.

This is all around gross. See https://github.com/facebook/flow/issues/267.

src/Root.js
danvk: No good reason -- I changed it to a blank. At some point there should be a clearer "loading" state.

src/GenomeTrack.js
danvk: Done

danvk:
Indeed it does! Addressed all comments, good to merge?

PR #16, opened 2015-02-26
Render colored base pairs using D3

danvk:
No dragging/panning or anything fancy, but it's a start!
image 2015-02-26 at 4 36 04 pm

### DIFF
--- a/playground.html
+++ b/playground.html
@@ -1,2 +1,20 @@
+<head>
+<meta http-equiv="Content-Type" content="text/html; charset=utf-8">
+<style>
+.reference {
+  width: 600px;
+  height: 20px;
+}
+.basepair.a { fill: #188712; }
+.basepair.g { fill: #C45C16; }
+.basepair.c { fill: #0600F9; }
+.basepair.t { fill: #F70016; }
+.basepair.u { fill: #F70016; }
+</style>
+</head>
+
+<body>
 <div id="root"></div>
+</body>
+
 <script src="build/all.js"></script>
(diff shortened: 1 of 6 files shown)


playground.html
```diff
@@ -1,2 +1,20 @@
...
+<style>
+.reference {
+  width: 600px;
+  height: 20px;
+}
+.basepair.a { fill: #188712; }
+.basepair.g { fill: #C45C16; }
+.basepair.c { fill: #0600F9; }
+.basepair.t { fill: #F70016; }
+.basepair.u { fill: #F70016; }
```
ihodes: Nothing to do for now, but it may be nice to use a CSS pre-processor at some pt, to make it much easier to skin the pileup viewer.

src/GenomeTrack.js
```diff
@@ -22,16 +23,69 @@ var GenomeTrack = React.createClass({
     }
 
     if (!this.props.basePairs) {
-      return <div>no data</div>;
+      return <div className="reference empty">no data</div>;
     }
 
-    return <div>{this.props.basePairs}</div>;
+    return <NonEmptyGenomeTrack key="ne" {...this.props} />;
```
ihodes: I don't think you need a key here?

ihodes:
Looks good!

playground.html
danvk: I'm hoping to use CSS for styling everywhere I can to enable this.

src/GenomeTrack.js
danvk: Good catch -- remnants of my missing return statement debugging session!

PR #33, opened 2015-03-13
Add Interval classes

danvk:
New modules:

- `Interval` represents a closed numeric interval.
- `ContigInterval` represents a closed interval on the genome (i.e. a contig + an `Interval`).
- `utils` contains a few helpers for ordering tuples, ala Python.

All of these are well-tested. None of them are used right now, but they'll all be helpful for working with bigBed files.

### DIFF
--- a/src/ContigInterval.js
+++ b/src/ContigInterval.js
@@ -1,0 +1,41 @@
+/* @flow */
+
+var Interval = require('./Interval');
+
+/**
+ * Class representing a closed interval on the genome: contig:start-stop.
+ *
+ * The contig may be either a string ("chr22") or a number (in case the contigs
+ * are indexed, for example).
+ */
+class ContigInterval {
+  contig: string|number;
+  interval: Interval;
+
+  constructor(contig: string|number, start: number, stop: number) {
+    this.contig = contig;
+    this.interval = new Interval(start, stop);
+  }
+
+  // TODO: make these getter methods & switch to Babel.
+  start() {
+    return this.interval.start;
+  }
+  stop() {
+    return this.interval.stop;
+  }
+  length() {
+    return this.interval.length();
+  }
+
+  intersects(other) {
+    return (this.contig == other.contig &&
+            this.interval.intersects(other.interval));
+  }
+
+  toString() {
+    return `${this.contig}:${this.start()}-${this.stop()}`;
+  }
+}
+
+module.exports = ContigInterval;
--- a/src/ContigInterval.js
+++ b/src/ContigInterval.js
@@ -18,22 +18,22 @@ class ContigInterval {
   }
 
   // TODO: make these getter methods & switch to Babel.
-  start() {
+  start(): number {
     return this.interval.start;
   }
-  stop() {
+  stop(): number {
     return this.interval.stop;
   }
-  length() {
+  length(): number {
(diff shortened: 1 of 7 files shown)


danvk:
If there are existing libraries for doing all this, I'd be happy to use them instead! A quick search on NPM didn't turn up anything quite like what I wanted—I think my `Interval` class is too simple for people to consider packaging it up.

src/ContigInterval.js
```diff
@@ -0,0 +1,41 @@
...
+class ContigInterval {
+  contig: string|number;
+  interval: Interval;
+
+  constructor(contig: string|number, start: number, stop: number) {
+    this.contig = contig;
+    this.interval = new Interval(start, stop);
+  }
+
+  // TODO: make these getter methods & switch to Babel.
```
ihodes: Sat more? (Not in the code, just curious!)

src/ContigInterval.js
```diff
@@ -0,0 +1,41 @@
...
+var Interval = require('./Interval');
+
+/**
+ * Class representing a closed interval on the genome: contig:start-stop.
+ *
+ * The contig may be either a string ("chr22") or a number (in case the contigs
+ * are indexed, for example).
+ */
+class ContigInterval {
+  contig: string|number;
```
ihodes: Would this make future code easier to deal with if contigs are always strings?

src/ContigInterval.js
danvk: According to the [es6features repo][1], this is supposed to be valid syntax:

```javascript
class Interval {
  get length() {
    return this.stop - this.start + 1;
  }
}
```

and then you can access `length` as though it were a property, not a method:

```javascript
new Interval(0, 10).length
```

I believe this is possible in ES5 using `Object.defineProperty`, but I don't want to muck with that. `react-tools` specifically doesn't support that syntax, but Traceur & Babel do.

[1]: https://github.com/lukehoban/es6features#classes

ihodes: Ah yeah, cool, makes sense. That's definitely possible in ES5 & it prevents the JIT from kicking, in just FYI.

src/ContigInterval.js
danvk: In bigBed format, they use a mapping which converts all the contig names to integers, for more compact storage. So it is convenient to allow `contig` to be either.

src/Interval.js
```diff
@@ -0,0 +1,68 @@
...
+    }
+    var result = intervals[0].clone();
+    intervals.slice(1).forEach(function({start, stop}) {
+      result.start = Math.max(start, result.start);
+      result.stop = Math.min(stop, result.stop);
+    });
+    return result;
+  }
+
+  static boundingInterval(intervals: Array<Interval>): Interval {
```
ihodes: Took me a minute to figure out what this was doing; maybe a comment?

danvk: Done

src/ContigInterval.js
danvk: > it prevents the JIT from kicking

as in it's bad for performance?

ihodes: Depending on how often it's called, it could be.

ihodes:
Cool, LGTM!

Issue #30, opened 2015-03-11
High-level API


danvk:
As described in the design doc, there should be a way to specify the track source and types in a JavaScript object.

danvk:
To start with, this might be as simple as passing in a list of (data source, visualization) tuples.

danvk:
I'm going to call this done! The API is defined enough for it to be used in CycleDash, at least.

PR #123, opened 2015-05-05
High level API

danvk:
**WIP**

This is a first crack at #30 

See `main.js` for an example of what the API looks like as of this PR.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/123)


danvk:
I'm going to go ahead and merge this.

All discussion happened on Slack. The consensus was to use user-generated objects for configuration, since this provides the maximum flexibility. No string→object/class mappings necessary. This will also allow me to swap in a new visualization API when I come up with it.

PR #124, opened 2015-05-07
Add {get,set}Location API methods

danvk:
I copied the name `setLocation` from Dalliance. I don't feel strongly about it, though, and I could be persuaded that there's a better name for it!

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/124)

### DIFF
--- a/src/Root.js
+++ b/src/Root.js
@@ -35,7 +35,6 @@ var Root = React.createClass({
     });
 
     referenceSource.on('contigs', () => {
-      // this is here to facilitate faster iteration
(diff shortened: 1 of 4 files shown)


ihodes:
Awesome—how does `{get,set}ViewableRange` or something with "range" sound? I'm also fine with location.

danvk:
What about just `{get,set}Range`? I'm not sure what other range this could be referring to.

ihodes:
That was my first inclination, too, wfm as well.

PR #143, opened 2015-05-15
Fixes & tests for PileupTrack.

### LINKED ISSUE #114: Genome and BAM tracks are off by one
Showing all the base pairs in reads make this clear:

screen shot 2015-04-30 at 3 44 44 pm

### LINKED ISSUE #115: Zooming out shows regions of all mismatches
I believe these correspond to regions where alignments are fetched before reads from the reference come in:

screen shot 2015-04-30 at 3 58 26 pm

Mismatches are computed once when the alignment comes in, whereas they should be updated whenver new reference data comes in.

danvk:
This PR includes a number of changes rolled together:

- Fixed a bug where zooming out could result in non-integer ranges, which broke `GenomeTrack`.
- TwoBit{,DataSource} are 0-based instead of 1-based
- `TwoBitDataSource` only fires `newdata` when new data comes in.
- `PileupTrack` recomputes alignment mismatches when the reference updates.
- Fixed some a subtle bug where the nested selection in `PileupTrack` was on the enter selection.
- Wrote tests for `PileupTrack` that force the alignments & reference data to come back in either order.
- Factored out `waitFor` async test helper.

I'd like to move to a world where all coordinate systems are 0-based, except for ones which are visible in the UI. Currently the D3 scale is 1-based, but I think it would be better if it were 0-based.

Fixes #114 
Fixes #115

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/143)


test/PileupTrack-test.js
```diff
@@ -0,0 +1,185 @@
...
+      expect(mismatches).to.have.length.below(20);
+      assertHasColumnOfTs();
+    });
+  });
+
+  // Same as the previous test, but with the loads reversed.
+  it('should indicate mismatches when the alignments load first', function() {
+    var {fakeTwoBit, fakeBam} = testSetup();
+
+    // Release the reference first.
```
ihodes: s/reference/alignment/

test/PileupTrack-test.js
```diff
@@ -0,0 +1,185 @@
...
+    fakeBam.release(alignments);
+
+    // Wait for the track to render, then release the alignments.
```
ihodes: s/alignments/references/

test/PileupTrack-test.js
danvk: Done

test/PileupTrack-test.js
danvk: Done

danvk:
Nice catch on the comments—fixed & will merge on green.

PR #173, opened 2015-06-08
Fix issue #172, infinite loop in BAM fetching

### LINKED ISSUE #172: Infinite loop in BAM fetching
See https://github.com/hammerlab/cycledash/issues/697

Fetching `chr20:2684600-2684800` in `http://hammerlab-dev3.hpc.mssm.edu:9876/datasets/martignetti_ovarian/189/Illumina_RNA/QC_S106.B202_PCT189_RNA.PE.RNASeqRibozero.RAPiD.Human/PT_189_Left_Ovary_RNA/Processed/RAPiD.2_0_0/bams/PT_189_Left_Ovary_RNA.accepted_hits.sort.coord.bam` throws the BAM fetcher into an infinite loop.

danvk:
Fixes #172 

Includes a regression test.

Previously, the logic on `bam.js:163-170` would produce a `newChunk` var identical to `blocks[0]` in the case that `blocks.length == 1`. Apparently this case only rarely comes up! I'm not totally confident that the issue is gone forever, but this particular instance is.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/173)


danvk:
@armish mind taking a look at this? I'm happy to show you what was going on in more detail if it doesn't make sense.

armish:
sure @danvk, let me try to replicate/isolate it and I will get you back if I cannot figure out the details.

armish:
@danvk: I see how these changes fix the problematic logic and the tests seem to be passing on my local machine as well (after a few refreshes :)) 

I think this is good to go :+1:

danvk:
Cool, thanks for taking a look!

PR #174, opened 2015-06-08
License identifier is now a valid SPDX expression

armish:
When installing the package, NPM was complaining about the license:

```
$ npm install
npm WARN package.json pileup@0.1.0 license should be a valid SPDX license expression
```

Changed the license description to `Apache-2.0`, for SPDX compatibility.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/174)

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "test": "grunt test"
   },
   "author": "Dan Vanderkam",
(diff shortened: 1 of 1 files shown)


danvk:
Nice!

PR #177, opened 2015-06-11
Rename the complete HTML demo to prevent name clashes across demos

armish:
This is to disambiguate naming clashes across demos that work with either *http-server* and *sniper*. Specifically, *sniper* will now know [http://localhost:9090/examples/playground](http://localhost:9090/examples/playground) resolves to [examples/playground.js](https://github.com/hammerlab/pileup.js/examples/playground.js).

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/177)


armish:
@danvk have you started hating all these BioJS changes, yet? :) I am hoping that we will not be dealing with these changes once this new version is out there.

PR #178, opened 2015-06-11
Add a demo using BioDalliance

danvk:
Now that we're using pileup.js on Cycledash, we don't have a reference for how dalliance renders BAM tracks. This adds one for comparison.

cc @jaclynperrone

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/178)


armish:
good to go on green :+1:

PR #175, opened 2015-06-11
Allow browserification via the standalone tool

armish:
The earlier configuration in `package.json` was confusing  [node-browserify](https://github.com/substack/node-browserify), forcing it to re-browserify `build/pileup.js` and therefore causing it to fail. As an example, see the output of [this browserify request](https://wzrd.in/bundle/pileup@0.1.2) for our version `v0.1.2`:

```
---FLAGRANT SYSTEM ERROR---

(logs uuid: 48fd78d0-1074-11e5-8345-05735e43db5c )

Error: "browserify exited with code 1"

code: 1
stderr: Error: Cannot find module './lib/utils/common' from '/tmp/pileup115511-14891-13ix8yo/node_modules/pileup/build'
    at /home/admin/browserify-cdn/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:42:25
    at load (/home/admin/browserify-cdn/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:60:43)
    at /home/admin/browserify-cdn/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:66:22
    at /home/admin/browserify-cdn/node_modules/browserify/node_modules/browser-resolve/node_modules/resolve/lib/async.js:21:47
    at Object.oncomplete (fs.js:107:15)

dirPath: /tmp/pileup115511-14891-13ix8yo

This is probably an issue with the package, and not browserify-cdn itself.
If you feel differently, feel free to file a bug report at:

    https://github.com/jfhbrook/browserify-cdn/issues

Include the ENTIRETY of the contents of this message, and the maintainer(s)
can try to help you out.

Have a nice day!
```

This error was manifesting itself within the demo on the [BioJS repository](http://workmen.biojs.net/demo/pileup/playground) through the [wzrd.in service](https://wzrd.in/) . I was able to track this problem down and found that `browser` field in our `package.json` was the reason for this particular error. This can also be replicated locally:

```
$ npm install pileup@0.1.2
pileup@0.1.2 node_modules/pileup
├── shallow-equals@0.0.0
├── jdataview@2.5.0
├── underscore@1.8.3
├── q@1.4.1
├── backbone@1.1.2
├── pako@0.2.7
├── jbinary@2.1.3 (es6-promise@2.3.0, request@2.57.0)
├── d3@3.5.5
└── react@0.13.3 (envify@3.4.0)

$ browserify node_modules/pileup
[...]

PR #187, opened 2015-06-12
Move jstransformify dependency from development to default

armish:
This is relevant to #175 (standalone browserification). The transform step of the browserify task requires jstransformify to be installed (globally or locally) before running `browserify`. The following replicates the issue:

```
$ cd /tmp
$ npm install pileup@0.1.3
pileup@0.1.3 node_modules/pileup
├── shallow-equals@0.0.0
├── jdataview@2.5.0
├── underscore@1.8.3
├── q@1.4.1
├── backbone@1.1.2
├── pako@0.2.7
├── jbinary@2.1.3 (es6-promise@2.3.0, request@2.57.0)
├── d3@3.5.5
└── react@0.13.3 (envify@3.4.0)
$ browserify --version
10.2.4
$ browserify node_modules/pileup
Error: Cannot find module 'jstransformify' from '/private/tmp/node_modules/pileup'
    at /usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:46:17
    at process (/usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:173:43)
    at ondir (/usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:188:17)
    at load (/usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:69:43)
    at onex (/usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:92:31)
    at /usr/local/lib/node_modules/browserify/node_modules/resolve/lib/async.js:22:47
    at FSReqWrap.oncomplete (fs.js:95:15)
```

and installing the `jstranformify` resolves the issue:

```
$ cd node_modules/pileup/
$ npm install git://github.com/danvk/jstransformify.git
$ browserify . > pileup.browserified.js && echo "browserified." || echo "failed."
browserified.
$ grep -i jstransform pileup.browserified.js && echo "jstansformify included" || echo "jstransformify not included"
jstransformify not included
```

and does not pull the `jstransformify` library into the bundle.

This also increases the minor revision by one to be able to re-deployed the package with the dependency fix.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/187)


danvk:
Is this for BioJS? I don't understand why `jstransformify` should be a non-dev dependency.. if a user does `require('pileup')`, they don't need to have `jstransformify` available.

armish:
This is not a BioJS requirement, but a `node-browserify` one. I had the jstranformify installed globally, so missed this problem with the `browserify` on my local. 

I agree with you that people doing a `require('pileup')` doesn't need jstransformify; but, playing devil's advocate: they actually don't need any of the default libraries either as we are simply distributing the browserified/bundled JS that has all of those libraries included in it.

And without transforming it, people actually cannot do `require('./src/pileup.js')` so I think `jstransformify` is a requirement but in a different sense.

danvk:
Then I guess I wouldn't expect your top example to work—users should be doing `require('pileup')`, not `require('pileup/src/pileup.js')`, right?

armish:
yeah, you are right. I think this PR simply addresses browserification issue but is actually not required for people who are not even going to bother with this step. Beside BioJS, this addition will also allow us to work nicely with [RequireBin](http://requirebin.com) service for generating demos from gists.

Currently, they don't like our bundling and cannot bundle on their own:
screen shot 2015-06-12 at 1 06 29 pm

danvk:
Would changing the `browser` field in `package.json` from `src/pileup.js` to `build/pileup.js` also fix the issue?

armish:
I first thought that should work, but turns out browserify is not smart enough to skip browserification if the file is already browserified. Our version `0.1.1` was like that and this is what you get:

https://wzrd.in/bundle/pileup@0.1.1
screen shot 2015-06-12 at 1 18 09 pm

Issue #180, opened 2015-06-11
The column containing your selected variant should be highlighted


jaclynperrone:
Highlighting the focused column will help orient the user. This behavior can be seen in IGV.

danvk:
i.e. the dotted vertical lines:

![screen shot 2015-06-11 at 6 53 49 pm](https://cloud.githubusercontent.com/assets/98301/8120263/3d17709c-106b-11e5-9c42-7f681239181c.png)

PR #195, opened 2015-06-16
Add track names

### LINKED ISSUE #129: Show track names
This is important for, say, knowing which pileup is Normal and which is Tumor.

IGV uses a left gutter ("Sequence", "RefSeq Genes"):
screen shot 2015-05-08 at 10 22 48 am

Dalliance shows them as overlaid semi-transparent bubbles ("DNase I"):
screen shot 2015-05-08 at 10 20 48 am

@jaclynperrone I'd love your thoughts on this! :)

danvk:
Here's what it looks like:

screen shot 2015-06-16 at 2 39 15 pm

This reworks `pileup.css` to use flexbox, which worked very well in Cycledash. You can see the shell of the layout in [this codepen][1].

@jaclynperrone any design feedback would be much appreciated here!

Fixes #129 

[1]: http://codepen.io/anon/pen/VLzbBe?editors=110

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/195)

### DIFF
--- a/style/pileup.css
+++ b/style/pileup.css
@@ -86,6 +86,7 @@
   stroke: blue;
 }
 .gene text {
+  font-family: 'Helvetica Neue', Helvetica, Arial, sans-serif;
   font-size: 16px;
   text-anchor: middle;
   stroke: black;
--- a/style/pileup.css
+++ b/style/pileup.css
@@ -16,6 +16,21 @@
 }
 .track-label {
   flex: 0 0 100px;  /* fixed-width track labels */
+  text-align: right;
+  font-family: Helvetica Neue, Helvetica, Arial, sans;
+  font-size: 0.9em;
+  position: relative;  /* make this an offset parent for positioning the label. */
+}
+.track-label > span {
+  padding-right: 5px;
+}
+/* bottom-justify these track labels */
+.track.reference .track-label > span,
+.track.variants .track-label > span
+{
+  position: absolute;
+  bottom: 0;
+  right: 0;
 }
 .track-content {
   flex: 1;  /* stretch to fill remaining space */
@@ -24,11 +39,12 @@
 /* controls */
 .pileup-root > .controls {
   flex: 0 0 30px;  /* fixed height */
+  margin-bottom: 0;  /* overrides browser default */
 }
 
 /* reference track */
 .pileup-root > .reference {
-  flex: 0 0 50px;  /* fixed height */
+  flex: 0 0 20px;  /* fixed height */
 }
 
 .background {
@@ -44,7 +60,7 @@
   text-anchor: middle;
 }
 .pair.loose text {
-  font-size: 28;
+  font-size: 24;
 }
 .pair.tight text {
   font-size: 12;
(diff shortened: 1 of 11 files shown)


style/pileup.css
```diff
@@ -4,11 +4,47 @@
...
+}
+
+.pileup-root > .track {
+  display: flex;
+  flex-direction: row;
+}
+.track-label {
+  flex: 0 0 100px;  /* fixed-width track labels */
+  text-align: right;
+  font-family: Helvetica Neue, Helvetica, Arial, sans;
```
jaclynperrone: Let's use the 'freight-sans' typekit font we use in Cycledash (unless you want to keep it universal)
`font-family: "freight-sans-pro", verdana, helvetica, sans-serif;`

If you want to keep it Helvetica:
Multi-word fonts are usually wrapped in quotation marks, and 'sans-serif' is used as the catch-all:
`font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;`

style/pileup.css
```diff
@@ -24,7 +60,7 @@
   text-anchor: middle;
 }
 .pair.loose text {
-  font-size: 28;
+  font-size: 24;
```
jaclynperrone: I think you may need a unit here? Preferably `em`

style/pileup.css
```diff
@@ -24,7 +60,7 @@
...
 }
 .pair.tight text {
   font-size: 12;
```
jaclynperrone: Same comment as above. Do you need to specify a unit?

style/pileup.css
danvk: Nice catch! I quoted correctly & changed `sans` to `sans-serif`.

I'd prefer to keep TypeKit out of core pileup.js. But users of pileup (like Cycledash) are free to use it in their own stylesheets.

style/pileup.css
danvk: Usually that would be true, but in this case I'm styling an SVG element. SVG has its own notion of coordinates, which are unit-free. This corresponds to 24px.

jaclynperrone:
Looking good! Added some line-notes. And a couple other little things:

- Can we have the gene name font be defaulted to Helvetica Neue as well? You could also get away with making it a little smaller. (And I'm debating whether or not it should be left-aligned.)
`font-family: "Helvetica Neue", Helvetica, Arial, sans-serif;`

- I would also try moving the gene name above the blue track. Right now it looks disconnected.

- The blue boxes in the variants panel are a little confusing to me. They look like they are supposed to contain something, but are showing up empty (simply adding a fill would fix this). They are also begging to be extended down through the pileup, kind of like how we talked about here #180. It would help keep a user's place as they scrolls through the BAM. What do you think? (Also sounds like it's for a different PR)

danvk:
I changed the gene name font to Helvetica Neue. I think the other items are for a follow-up PR.

The gene track could definitely benefit from some design thinking. I tried to mirror IGV, which centers the gene name under the portion of the transcript that's visible, but it doesn't make as much sense visually w/o borders separating all the tracks.

jaclynperrone:
Yes! I forgot to mention the borders thing. It would be good to use borders to separate the tracks. But yes, all these things can be for another time.

PR #203, opened 2015-06-22
Fire progress events for BAI fetching

danvk:
Fetching the BAI file can be a significant part of the load time for pileup tracks, especially if there are no index chunks available. This will give more immediate feedback to the user that activity is happening.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/203)

### DIFF
--- a/src/bam.js
+++ b/src/bam.js
@@ -207,11 +207,17 @@ class Bam {
 
     var sizePromise = this.index ? this.index.getHeaderSize() : Q.when(2 * 65535);
     this.header = sizePromise.then(size => {
-      return this.remoteFile.getBytes(0, size).then(buf => {
-        var decomp = utils.inflateGzip(buf);
-        var jb = new jBinary(decomp, bamTypes.TYPE_SET);
-        return jb.read('BamHeader');
-      });
+      var def = Q.defer();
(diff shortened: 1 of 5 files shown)


src/bam.js
```diff
@@ -294,10 +300,18 @@ class Bam {
     var index = this.index;
 
     return this.getContigIndex(range.contig).then(({idx, name}) => {
+      var def = Q.defer();
+      // This happens in the next event loop to give listeners a chance to register.
```
ihodes: Oh boy

src/utils.js
```diff
@@ -150,6 +152,21 @@ function altContigName(contig: string): string {
   }
 }
 
+/**
+ * Pipe all promise events through a deferred object.
+ * This is similar to deferred.resolve(promise), except that it allows progress
```
ihodes: I am definitely not a promises expert, but this seems different than `.resolve()`? It's just adding a `then` to the deferred?

danvk: I believe this should be exactly the same as `deferred.resolve(promise)`, which resolves/rejects the deferred with whatever the promise eventually winds up as. Except for the notifications part.

PR #204, opened 2015-06-25
Hacky workaround for #200

danvk:
Fixes #200 

We don't care much about these errors, so we may as well suppress them while we work on a better solution.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/204)

Issue #200, opened 2015-06-22
"Required module not found" Flow errors


winkler1:
It's really cool that you're using Flow so heavily in this. Getting errors though--

$ grunt travis
Running "flow:app" (flow) task
>> /Users/Jeff/repos/ghtemp/src/BamDataSource.js:12,8,11: Unexpected token type
>> 
>> /Users/Jeff/repos/ghtemp/src/BigBed.js:16,22,48: ContigInterval.js
>> Required module not found
>> 
>> /Users/Jeff/repos/ghtemp/src/BigBedDataSource.js:9,22,48: ContigInterval.js
>> Required module not found
>> 
>> /Users/Jeff/repos/ghtemp/src/ContigInterval.js:12,23,23: Unexpected token :
>> 
>> /Users/Jeff/repos/ghtemp/src/Controls.js:25,35,48: identifier SyntheticEvent
>> Unknown global name
>> 
>> /Users/Jeff/repos/ghtemp/src/Controls.js:39,23,61: call of method indexOf
>> Method cannot be called on possibly null or undefined value
>>   /Users/Jeff/repos/ghtemp/src/Controls.js:14,17,63: ?arrayOf
>> 
>> /Users/Jeff/repos/ghtemp/src/Controls.js:66,11,85: call of method map
>> Method cannot be called on possibly null or undefined value
>>   /Users/Jeff/repos/ghtemp/src/Controls.js:14,17,63: ?arrayOf

danvk:
@winkler1 thanks for trying pileup.js! Can you check which version of Flow you're using?

The three "Required module not found" errors are a known issue. @armish is there an easy workaround available for this before we address #188? We're in a state that's a bit confusing for new users.

I only see three errors when I run `grunt travis`, not 81!

```
grunt travis
Running "flow:app" (flow) task
>> /Users/danvk/github/pileup.js/test/GenomeTrack-test.js:14,14,30: pileup
>> Required module not found
>>
>> /Users/danvk/github/pileup.js/test/PileupTrack-test.js:16,14,30: pileup
>> Required module not found
>>
>> /Users/danvk/github/pileup.js/test/components-test.js:6,14,30: pileup
>> Required module not found
>>
>> Found 3 errors
Warning: Task "flow:app" failed. Use --force to continue.

Aborted due to warnings.
```

winkler1:
[...]

Issue #188, opened 2015-06-12
Only publish tranformed Javascript code to NPM


armish:
This enhancement option came up in the context of issues #187 and #175. Specifically, the problem is that publishing non-transformed code causes problems for people building on top of our library and its automated `browserification` (and integration) is extremely problematic.

We currently transform and browserify the code via `grunt prod` and publish the bundled version for people to use on their websites. #175 was an attempt to make the project compatible with the *de facto* browserify tool, but it required us to introduce new dependencies into the project, which was not ideal.

Here is a relevant question (and a workaround to these issues) on StackOverflow: [Can I write npm package in coffeescript?](http://stackoverflow.com/questions/13645824/can-i-write-npm-package-in-coffeescript). It would be great to refactor pileup to also use this workflow, where we pre-transform the code and only publish the Javascript code to NPM for better compatibility for other people.

danvk:
Summarizing a discussion @armish and I just had:

- We'll move the existing `src` and `test` directories under a top level `src` directory (`src/src`? Maybe just move the `test` directory under `src`).
- We'll transform the code file-by-file into the `build` directory, which will only contain plain JavaScript.

Things we need to work after this change:

- [ ] Running `grunt flow` returns no errors (currently [returns 3][1])
- [ ] Running `grunt test` passes (i.e. running the unit tests under Phantom)
- [ ] Running `grunt browserify:watchTest` and loading the tests in-browser shows everything passing.
- [ ] Installing pileup.js as an NPM package and then doing `var pileup = require("pileup")` results in a functional pileup viewer (e.g. in Cycledash)
- [ ] Sourcing `pileup.min.js` results in a global `pileup` object for those who don't want to use Browserify.

[1]: https://github.com/hammerlab/pileup.js/issues/200

PR #205, opened 2015-06-25
Refactor code for better compatibility with CommonJS, Flow, Browserify and Watchify

armish:
These changes 

- collect the application and test code files under a unified directory: `src/{main, test}`
- add steps that, by default, **jstransform** and **browserify** the code from `src` into `build` (issue #188)
- modify `package.json` so that the **main** and **browser** fields now point to plain JS-code under `build`
- drops aliasing within **grunt-browserify** so that both watched tests and flow checks run smoothly (issue #200 and PR #191)
- reverts the hacks introduced by PR #204)

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/205)


danvk:
Thanks for making this huge change! It's going to simplify distribution immensely.

I've left a few scattered comments, but my main bit of feedback is that we should either ditch Grunt in favor of shell scripts/npm scripts or use it for everything. I don't want two build systems. I'm fine going either way. Arguments in favor of grunt are that all the commands are centralized and JavaScript is a better language than shell. Arguments in favor of shell scripts are that we can ditch the `grunt-*` wrappers around everything we want to use.

---

Reviewed 9 of 185 files at r1.
Review status: 9 of 185 files reviewed at latest revision, 16 unresolved discussions, all commit checks successful.

---

<sup>**[.flowconfig, line 9 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskNj5dmOE9-qL3A86i)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/.flowconfig#L9)):</sup>
I believe this is superfluous -- everything under the project root will be included by default. See http://flowtype.org/docs/advanced-configuration.html#example

---

<sup>**[examples/playground-complete.html, line 19 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskRD0X3gis4juJ_lDt)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/examples/playground-complete.html#L19)):</sup>
Any reason for the rename? Since the name of the product is "pileup.js", `pileup.js` seems like a good name for the JS bundle :)

I initially read this as "pileup browser", rather than "pileup built for the web browser".

---

<sup>**[examples/playground.js, line 1 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskROqrxXOPIEcVB2Gh)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/examples/playground.js#L1)):</sup>
This has always puzzled me—when you source `pileup.min.js`, does it define `require` globally?  That is somewhat surprising behavior.

---

<sup>**[examples/playground.js, line 29 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskRlqaRCxZOiqJC9UQ)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/examples/playground.js#L29)):</sup>
Why this reorganization? (`/test/data`→`/test-data`)

---

<sup>**[package.json, line 20 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskXu7LYJl9josHUAEq)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/package.json#L20)):</sup>
Looking at other packages, it seems like `dist` is a bit more standard (jQuery and react use it, for example). The other common pattern is to put all the build artifacts in a top-level directory (like Q and D3).

Renaming to `dist` is probably easier. What do you think? I know I was the one who suggested `build`—sorry!

---

<sup>**[package.json, line 21 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskXdCQXifjNmPOB_qJ)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/package.json#L21)):</sup>
I don't think we want to distribute `src` any more: users will only be interested in the plain-old-JavaScript version.

---

<sup>**[package.json, line 29 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskSUEhyR7muc8Lwy6v)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/package.json#L29)):</sup>
Do you want to abandon grunt?

---

<sup>**[README.md, line 13 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskSmSFapL128ylrMEu)** ([raw file](https://github.com/hammerlab/pileup.js/blob/ebfb445487da7cad593fa85938d7d02a01cf6bc5/README.md#L13)):</sup>
This will only work if you have `node_modules/.bin` on your `$PATH`. Maybe worth calling that out, or leaving as-is.

---

<sup>**[README.md, line 14 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskT43dEMXRhSJ2paJk)** ([raw file](https://github.com/hammerlab/pileup.js/blob/ebfb445487da7cad593fa85938d7d02a01cf6bc5/README.md#L14)):</sup>
You don't need to build anything?

We should reorganize these sections. The "Usage" section is more interesting to users and should go on top. The "Quickstart" section is of interest to developers working on pileup and should be renamed something like "Development Setup".

---

<sup>**[README.md, line 96 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskU1TGPUA96hmM_WS2)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/README.md#L96)):</sup>
Are both of these commands necessary?

---

<sup>**[README.md, line 98 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskTsC4wolNxMmRMGPL)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/README.md#L98)):</sup>
The quotes aren't necessary here.

---

<sup>**[README.md, line 107 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskTXL_1BjFR3isYMjj)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/README.md#L107)):</sup>
It's actually way better to run `flow status` for type checking. The `grunt flow` command starts up a new Flow process and type checks everything from scratch, which takes 10-20 seconds. `flow status` starts a background server which takes 10-20 seconds the first time, but ~100ms thereafter.

---

<sup>**[scripts/jstransform.sh, line 3 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskUFS0EcpYxUbgWT_i)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/scripts/jstransform.sh#L3)):</sup>
you don't need to `export` this variable to the environment—you only use it in this script, rather than in child processes.

---

<sup>**[scripts/jstransform.sh, line 3 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskUXN6122XMIoZPuOL)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/scripts/jstransform.sh#L3)):</sup>
Change `${JSTRANSFORM_OPTS}` to `$@` and move it to the end. That way you can invoke this script as `./scripts/jstransform.sh --watch`, rather than dealing with environment variables. It will also let you override default options, e.g. if you wanted to do `./scripts/jstransform.sh --noharmony`:

```bash
JSTRANSFORM_OPTS="--react --harmony --strip-types --non-strict-es6module --source-map-inline $@"
```

---

<sup>**[scripts/jstransform.sh, line 4 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskUCgz1WweubM2Imgq)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/scripts/jstransform.sh#L4)):</sup>
remove debug code

---

<sup>**[scripts/jstransform.sh, line 5 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskV3MlRrgM-r0wjRUQ)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/scripts/jstransform.sh#L5)):</sup>
Move this comment up to the top (after the shebang line). It describes what the entire script does.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/205)

armish:
Thanks for the review, Dan! I hear you about the grunt-vs-bash dichotomy. I also like grunt and JS-based tasks much better than the scattered bash scripts -- but that thing about grunt wrappers was a huge barrier for me. I wanted to separate the `jstransform` and `browserify` steps from each other, but there was really no good `grunt-jstransform` wrapper out there. It should be easy to simply create one, but I don't want to go down that path at this point.

I think in this situation I am leaning more towards bash/NPM scripts rather than grunt as it is the simplest solution (not the most beautiful one, though) towards our goal. We currently depend on Grunt for uglifying and flow checks, but those two can easily be replaced with their script-based counterparts, which I will do if you agree.

---

Review status: 9 of 185 files reviewed at latest revision, 15 unresolved discussions, some commit checks failed.

---

<sup>**[.flowconfig, line 9 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskNj5dmOE9-qL3A86i)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/.flowconfig#L9)):</sup>
agree that this is not necessary -- just wanted to put it there to emphasize that we are flow-checking the code under `src`.

---

<sup>**[examples/playground-complete.html, line 19 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskRD0X3gis4juJ_lDt)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/examples/playground-complete.html#L19)):</sup>
While I was thinking about this library in general, I thought that this can potentially be used as a backend library to fetch/parse BAM/VCF files without visualization on the browser. `pileup.js` is concise and clean, but if we would like to create bundled versions also for backend use, that might cause some confusion. Thinking about this for a second time, I think it is better we stick to the short version `pileup.js` for what it is worth.

Will fix this.

---

<sup>**[examples/playground.js, line 1 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskROqrxXOPIEcVB2Gh)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/examples/playground.js#L1)):</sup>
only if it is not defined previously. So all browserified libraries first check whether `require` is already defined and if so it simply modifies it to recognize the new package names. Otherwise they define a global `require` function by themselves. You can find that logic in the first few lines of the browserified source code for all these libraries.

This was also confusing to me as this is not compatible with node's require; but amazing enough, when you have multiple browserified libraries, it works just fine :)

---

<sup>**[examples/playground.js, line 29 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskRlqaRCxZOiqJC9UQ)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/examples/playground.js#L29)):</sup>
that was because I moved all the test source files over to a different directory. I thought that having a `test` folder with only a single sub-directory was not necessary. The alternative was to still keep these files under `src/test/data` but they don't survive the `jstransform` and it would have meant to require use of longer paths in test/examples.

I think this is a clean solution, where we have a single folder for all the example data -- it is also easier to refer them from the source code.

---

<sup>**[package.json, line 20 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskXu7LYJl9josHUAEq)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/package.json#L20)):</sup>
no worries, I would love to get this right while we are at it. Dist makes sense, but does that mean we are going to completely get rid of the `build` directory? 

---

<sup>**[package.json, line 21 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskXdCQXifjNmPOB_qJ)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/package.json#L21)):</sup>
That is correct -- my only worry is that if we don't distribute `src`, then people doing 

```
npm install --save pileup
cd pileup
npm install
```

will get an error about the default `jstransform` step as they won't have the original untransformed files; but I don't think it is common practice to download the NPM module for development purposes from the NPM repository, right? If so, yeah -- we can exclude `src` from the distribution.

---

<sup>**[package.json, line 29 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskSUEhyR7muc8Lwy6v)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/package.json#L29)):</sup>
I think so (also see my overall comment on this)

---

<sup>**[README.md, line 13 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskSmSFapL128ylrMEu)** ([raw file](https://github.com/hammerlab/pileup.js/blob/ebfb445487da7cad593fa85938d7d02a01cf6bc5/README.md#L13)):</sup>
I didn't know this, but when you run scripts via `npm run`, npm appends all `node_modules/*/bin` to your path. But this reminds me that I forgot to add `grunt-cli` to the dev dependencies for this to work.

---

<sup>**[README.md, line 14 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskT43dEMXRhSJ2paJk)** ([raw file](https://github.com/hammerlab/pileup.js/blob/ebfb445487da7cad593fa85938d7d02a01cf6bc5/README.md#L14)):</sup>
the builds are now part of the default `npm install`. Apparently when you override the `install` script within `package.json`, you can add extra steps to the default install procedure. I used this trick to make the build as part of the installation.

Agree about re-organizing the README. I will do that.

---

<sup>**[README.md, line 96 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskU1TGPUA96hmM_WS2)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/README.md#L96)):</sup>
The first one watches the untransformed files and the second one watches the transformed files. To propagate changes in the untransfomed files to the tests, you need these two watchified commands. 

I tried to look for solutions that enable watchifying both, but wasn't able to find one. I might be able to collapse these two steps into a single script to make things easier, though -- e.g., `./scripts/watchTests.sh`. Do you think this is a good idea?

---

<sup>**[README.md, line 98 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskTsC4wolNxMmRMGPL)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/README.md#L98)):</sup>
I try to quote them to force the syntax highlighters to treat it as a string and highlight them using a single color, but sure -- I can get rid of the quotes.

---

<sup>**[README.md, line 107 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskTXL_1BjFR3isYMjj)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/README.md#L107)):</sup>
It might be my setup, but `flow status` was not able to start the server by default on my machine. If it the `status` command is faster, we should definitely use that instead. Let me double-check thins on my laptop and fix this then.

---

<sup>**[scripts/jstransform.sh, line 3 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskUXN6122XMIoZPuOL)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/scripts/jstransform.sh#L3)):</sup>
that is a good idea -- will do that.

---

<sup>**[scripts/jstransform.sh, line 4 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskUCgz1WweubM2Imgq)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/scripts/jstransform.sh#L4)):</sup>
argh! :\

---

<sup>**[scripts/jstransform.sh, line 5 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskV3MlRrgM-r0wjRUQ)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/scripts/jstransform.sh#L5)):</sup>
:thumbsup:

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/205)

danvk:
I'm fine with ditching Grunt in favor of shell scripts. Make it so :)

---

Review status: 9 of 185 files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

<sup>**[.flowconfig, line 9 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskNj5dmOE9-qL3A86i)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/.flowconfig#L9)):</sup>
Let's drop it, then—reading this config file, you'd think it was necessary, or that it limits Flow to code in `src` (which I don't think it does).

---

<sup>**[examples/playground.js, line 1 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskROqrxXOPIEcVB2Gh)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/examples/playground.js#L1)):</sup>
That's cool, thanks for the explanation!

---

<sup>**[package.json, line 20 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskXu7LYJl9josHUAEq)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/package.json#L20)):</sup>
Yes, we'd just rename the existing `build` directory to `dist`.

---

<sup>**[package.json, line 21 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskXdCQXifjNmPOB_qJ)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/package.json#L21)):</sup>
Yeah, that's not a mode I care about supporting. If you want to depend on pileup, you can install it and either `<script src>` from `node_modules/pileup` or `require('pileup')`. If you want to develop pileup.js, you can clone the git repo.

---

<sup>**[README.md, line 13 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskSmSFapL128ylrMEu)** ([raw file](https://github.com/hammerlab/pileup.js/blob/ebfb445487da7cad593fa85938d7d02a01cf6bc5/README.md#L13)):</sup>
That's good to know!

---

<sup>**[README.md, line 96 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskU1TGPUA96hmM_WS2)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/README.md#L96)):</sup>
Yes, I think that would be helpful!

---

<sup>**[README.md, line 107 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskTXL_1BjFR3isYMjj)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/README.md#L107)):</sup>
Yeah, let's try to get `flow status` working on your machine. It's going to be a life saver when you're working on features.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/205)

armish:
Reviewed 183 of 185 files at r1, 2 of 2 files at r2.
Review status: 174 of 190 files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/205)

danvk:
This looks great! I have just a few minor nits & then we'll be good to merge.

---

Review status: 174 of 190 files reviewed at latest revision, 8 unresolved discussions, all commit checks successful.

---

<sup>**[.flowconfig, line 9 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskNj5dmOE9-qL3A86i)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/.flowconfig#L9)):</sup>
Still suggest dropping this -- it's intended for including code in directories outside of the main project.

---

<sup>**[.travis.yml, line 6 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-Jt57Ys3KpLqoBRPZZQs)** ([raw file](https://github.com/hammerlab/pileup.js/blob/1080f42a51dcf5abe7d93a2eac7ad0d4b900aae8/.travis.yml#L6)):</sup>
`npm run lint` ?

---

<sup>**[package.json, line 32 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-Jt59O5Fa-UD0RviX8oB)** ([raw file](https://github.com/hammerlab/pileup.js/blob/1080f42a51dcf5abe7d93a2eac7ad0d4b900aae8/package.json#L32)):</sup>
It's a bit surprising that `scripts/watch.sh` and `npm run watch` do different things. Do you want to call this something like `watch-dist`? Or make it run the script?

---

<sup>**[README.md, line 102 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-Jt57z2alOWz0mYYGDUm)** ([raw file](https://github.com/hammerlab/pileup.js/blob/1080f42a51dcf5abe7d93a2eac7ad0d4b900aae8/README.md#L102)):</sup>
`npm watch-test`?

---

<sup>**[scripts/jstransform.sh, line 4 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-Jt58RQKdAVCQBLxtxQt)** ([raw file](https://github.com/hammerlab/pileup.js/blob/1080f42a51dcf5abe7d93a2eac7ad0d4b900aae8/scripts/jstransform.sh#L4)):</sup>
Move each flag onto its own line -- there's no significance to the grouping here.

---

<sup>**[scripts/jstransform.sh, line 6 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-Jt58GuWrAANHiHE_zFi)** ([raw file](https://github.com/hammerlab/pileup.js/blob/1080f42a51dcf5abe7d93a2eac7ad0d4b900aae8/scripts/jstransform.sh#L6)):</sup>
This may need to be `"$@"`, see https://www.biostars.org/p/18642/

---

<sup>**[scripts/test.sh, line 7 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-Jt58cR5VM_2l_QheZ10)** ([raw file](https://github.com/hammerlab/pileup.js/blob/1080f42a51dcf5abe7d93a2eac7ad0d4b900aae8/scripts/test.sh#L7)):</sup>
Was this actually necessary in practice?

---

<sup>**[scripts/test.sh, line 7 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-Jt58eKHB2SH_h3L_vCL)** ([raw file](https://github.com/hammerlab/pileup.js/blob/1080f42a51dcf5abe7d93a2eac7ad0d4b900aae8/scripts/test.sh#L7)):</sup>
Style nit: we typically put two spaces between end of line an inline comment.

---

<sup>**[scripts/watch.sh, line 2 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-Jt58vvArITRjE-sZPvU)** ([raw file](https://github.com/hammerlab/pileup.js/blob/1080f42a51dcf5abe7d93a2eac7ad0d4b900aae8/scripts/watch.sh#L2)):</sup>
effect → affect

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/205)

armish:
Review status: 174 of 190 files reviewed at latest revision, 6 unresolved discussions, all commit checks successful.

---

<sup>**[.flowconfig, line 10 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-JskNj5dmOE9-qL3A86i)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bb685cc6b09a50162f054ca7bdffe7e787399dc3/.flowconfig#L10)):</sup>
my bad -- I actually dropped this, but started getting flow errors at the last minute; thought this will solve the problem, but it didn't and apparently forgot it up there :\

---

<sup>**[package.json, line 32 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-Jt59O5Fa-UD0RviX8oB)** ([raw file](https://github.com/hammerlab/pileup.js/blob/1080f42a51dcf5abe7d93a2eac7ad0d4b900aae8/package.json#L32)):</sup>
yepp, it makes sense. We now have `npm run watch`, which calls `./scripts/watch.sh` and the distribution watch task is now called `watch-dist` as you suggested.

---

<sup>**[README.md, line 102 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-Jt57z2alOWz0mYYGDUm)** ([raw file](https://github.com/hammerlab/pileup.js/blob/1080f42a51dcf5abe7d93a2eac7ad0d4b900aae8/README.md#L102)):</sup>
this is now `npm run watch` (watches both the test and dist files).

---

<sup>**[scripts/jstransform.sh, line 6 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-Jt58GuWrAANHiHE_zFi)** ([raw file](https://github.com/hammerlab/pileup.js/blob/1080f42a51dcf5abe7d93a2eac7ad0d4b900aae8/scripts/jstransform.sh#L6)):</sup>
oh, good catch -- thanks!

---

<sup>**[scripts/test.sh, line 7 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-Jt58cR5VM_2l_QheZ10)** ([raw file](https://github.com/hammerlab/pileup.js/blob/1080f42a51dcf5abe7d93a2eac7ad0d4b900aae8/scripts/test.sh#L7)):</sup>
actually no -- but it was just my gut feeling that we might run into random racing conditions in the future. I will mention this in the comments for future reference.

---

<sup>**[scripts/watch.sh, line 2 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-Jt58vvArITRjE-sZPvU)** ([raw file](https://github.com/hammerlab/pileup.js/blob/1080f42a51dcf5abe7d93a2eac7ad0d4b900aae8/scripts/watch.sh#L2)):</sup>
:thumbsup: 

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/205)

danvk:
Review status: 174 of 190 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[scripts/test.sh, line 12 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-Jt58cR5VM_2l_QheZ10)** ([raw file](https://github.com/hammerlab/pileup.js/blob/1080f42a51dcf5abe7d93a2eac7ad0d4b900aae8/scripts/test.sh#L12)):</sup>
Let's remove it until we see it be an issue. I'm guessing that, in practice, `npm run mocha-phantomjs` has at least as much startup overhead as `http-server`.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/205)

danvk:
Just one more comment—feel free to merge on green!

---

Review status: 174 of 190 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/205)

armish:
Thanks for the reviews, Dan! Just addressed your comment in the latest revision.

---

Review status: 174 of 190 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

<sup>**[scripts/test.sh, line 12 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/205#-Jt58cR5VM_2l_QheZ10)** ([raw file](https://github.com/hammerlab/pileup.js/blob/1080f42a51dcf5abe7d93a2eac7ad0d4b900aae8/scripts/test.sh#L12)):</sup>
I agree -- I think I am being over-cautious on this one. I will remove the sleep step.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/205)

danvk:
Great! Let's merge on green.

---

Review status: 174 of 190 files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/205)

armish:
argh! Looks like mocha-phantomjs is faster than http-server :\

---

Review status: 174 of 190 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/205)

danvk:
OK, sleep it is :)

---

Review status: 174 of 190 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/205)

armish:
undo'ed that last comment. Merging this PR in, but we probably need to bump the version up soon.

---

Review status: 174 of 190 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/205)

danvk:
Yep, we should do a version bump once this goes in and we see it working in Cycledash!

PR #206, opened 2015-06-26
Add test and fix js/css for control buttons.

jaclynperrone:
This commit does a few things for the control button styles:

1. Changes `class` to `className` in `Controls.js`
2. Adds test to look for `.zoom-controls` div
3. Fixes css to display styles correctly

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/206)


danvk:
LGTM

Ideally we'd catch the `class` vs `className` issue via a linter — see https://github.com/STRML/JSXHint/issues/26

Since it did log a warning on the console, we might be able to catch this as part of the tests. I've filed #207 for that.

PR #210, opened 2015-06-30
Throw on warnings during tests

### LINKED ISSUE #207: React warnings should make the tests fail
React warns about issues like `class` vs. `className` and `for` vs `htmlFor` in the console. We should catch these when running the tests and make them failures.

```
$ grunt test
  GenomeTrack
Warning: Unknown DOM property class. Did you mean className?
Fetching 60 base pairs
    ✓ should tolerate non-chr ranges (212ms)
    ✓ should zoom in and out (157ms)
```

danvk:
Fixes #207

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/210)


danvk:
I'm going to abandon this. For some reason the tests aren't surfacing React warnings about `class` vs `className`, which was the whole point of this change.

PR #214, opened 2015-07-01
publish script

danvk:
This gets the `dist` directory into a good state for publication on npm.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/214)

### DIFF
--- a/scripts/publish.sh
+++ b/scripts/publish.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+# Publish an updated version of pileup.js on NPM.
+
+set +x
+set -o errexit
+
+# Ensure a clean dist directory
+rm -rf dist
+
+npm run jstransform
+npm run browserify
+npm run uglify
+
+# jstransform creates this, but we don't want to distribute it.
+rm -rf dist/.module-cache
+
+# No need to distribute tests
+rm -rf dist/test
+
+# Once we have more confidence in it, this script should just run `npm publish`.
+echo
+echo "Now run npm publish"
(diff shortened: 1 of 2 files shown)


scripts/publish.sh
```diff
@@ -0,0 +1,22 @@
...
+
+npm run jstransform
+npm run browserify
+npm run uglify
+
+# jstransform creates this, but we don't want to distribute it.
+rm -rf dist/.module-cache
+
+# No need to distribute tests
+rm -rf dist/test
```
armish: how about also removing the browserified test file? 
```
rm -f dist/tests.js
```

scripts/publish.sh
```diff
@@ -0,0 +1,22 @@
...
+
+set +x
+set -o errexit
+
+# Ensure a clean dist directory
+rm -rf dist
+
+npm run jstransform
+npm run browserify
+npm run uglify
```
armish: if you'd like, you can shorten these three runs as: `npm run build` (up to you, not really necessary)

scripts/publish.sh
danvk: That doesn't get generated because I run `npm run browserify`, rather than `npm run browserify-all`.

scripts/publish.sh
danvk: That would run `browserify-all` though, which I don't want (see below) :)

armish:
looks good, I have two minor comments (feel free to accept/reject them)

scripts/publish.sh
armish: :)

PR #216, opened 2015-07-02
Style fixes

### LINKED ISSUE #208: CSS issues
A few CSS issues that have popped up:

- [x] +/- buttons are wrapped to the next line
- [ ] Pileup tracks are too tall by default -- they seem to have their height set to the page's height, not their content's height.
- [ ] Scrolling a pileup track scrolls its label out of view. Only the content should scroll.

jaclynperrone:
This commit does a couple things:

- Closes #208 
- Styles loading indicator

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/216)

### DIFF
--- a/style/pileup.css
+++ b/style/pileup.css
@@ -35,6 +35,7 @@
 .track-content {
   flex: 1;  /* stretch to fill remaining space */
   overflow-y: auto;
+  overflow-x: hidden;
   position: relative;
 }
 .track-content > div > div {
@@ -44,6 +45,8 @@
 /* controls */
 .pileup-root > .controls {
   flex: 0 0 30px;  /* fixed height */
+}
+.pileup-root > .controls form.controls {
   margin-bottom: 0;  /* overrides browser default */
 }
 .pileup-root > .controls .zoom-controls {
@@ -124,7 +127,6 @@
 /* pileup track */
 .pileup-root > .pileup {
   flex: 1;  /* stretch to fill remaining space */
-  border-top: 1px solid #666;
 }
 .pileup .alignment .match {
   fill: #c8c8c8;  /* matches IGV */
--- a/style/pileup.css
+++ b/style/pileup.css
@@ -34,6 +34,11 @@
 }
 .track-content {
   flex: 1;  /* stretch to fill remaining space */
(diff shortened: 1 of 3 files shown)


style/pileup.css
```diff
@@ -119,7 +124,7 @@
 /* pileup track */
 .pileup-root > .pileup {
   flex: 1;  /* stretch to fill remaining space */
-  overflow-y: auto;
+  border-top: 1px solid #666;
```
danvk: I wonder if we should add this to all tracks? Putting a 1px border on top of pileup tracks is specific to the arrangement that we're using on the demo page, but it might not make sense if the pileup track were on top, say, and the other tracks were underneath it.

style/pileup.css
```diff
@@ -138,11 +143,20 @@
   stroke-width: 2;
 }
 .pileup .network-status {
-  position: absolute;
-  font-size: small;
+  height: 100%;
```
danvk: Can you include a screenshot of what this looks like? (I know, I know, we should have screenshot tests for pileup.)

danvk:
I'm seeing vertical scrollbars on tracks other than the pileups, and horizontal scrollbars on either the pileup tracks (when the page loads) or the other tracks (when I resize):

Ideally we'd never have horizontal scroll bars and we'd only have vertical scrollbars on the pileup tracks when they're necessary.

scrollies

jaclynperrone:
Ahh, weird. I don't see those scrollbars in my browser. I'll take a look

jaclynperrone:
Screenshot of the loading indicator
screen shot 2015-07-06 at 12 05 13 pm

style/pileup.css
jaclynperrone: Maybe we should just not have a border? Adding it to all the tracks looks a little excessive (as seen below). Now that the labels stay put as you scroll, you may be able to get away adding some margin between the tracks instead. What do you think?
border

danvk: I agree that borders on every track looks a bit excessive. How does the margin approach look? I happily defer to your judgment on this one :)

danvk:
This looks great! Feel free to merge at your leisure.

PR #217, opened 2015-07-07
Run build before running test

danvk:
We're not actually running the tests on Travis right now. mocha-phantomjs just says "0/0 passed", which counts as success:

https://travis-ci.org/hammerlab/pileup.js/builds/69762743

This is one thing we lose by dropping Grunt -- we need to track dependencies between build steps in some other way.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/217)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,4 +3,8 @@ language: node_js
 node_js:
   - "0.12"
 
-script: npm run test && npm run flow-check && npm run lint
+script: >
+    npm run build &&
+    npm run test &&
+    npm run flow-check &&
+    npm run lint


package.json
```diff
@@ -78,6 +78,7 @@
     "es5-shim": "^4.1.0",
     "flow-bin": "^0.12.0",
     "http-server": "^0.8.0",
+    "istanbul": "^0.3.17",
```
armish: is` istanbul` required for the tests to run on Travis?

danvk: No, reverting... :)

armish:
good to go on green :+1: 

thanks for catching this.

PR #220, opened 2015-07-07
Add an FPS meter to playground-complete.html

danvk:
This should help us keep track of frame rates as we add features/optimizations. It seems to correlate well with how things feel as I pan back and forth. If I zoom out a few times and pan, I can see the frame rate drop to 1 or 2 fps.

fps meter

If you click the meter, it switches between ms/frame and fps mode.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/220)

### DIFF
--- a/examples/playground-complete.html
+++ b/examples/playground-complete.html
@@ -6,8 +6,13 @@
   position: absolute;
   left: 10px;
   right: 10px;
-  top: 10px;
+  top: 20px;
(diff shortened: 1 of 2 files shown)


armish:
cute widget -- didn't know such a thing existed or you could use it to measure performance.

Feel free to merge it :+1:

PR #222, opened 2015-07-08
Tweaks to the test script

danvk:
- I noticed that whenever I ran `npm run test`, it left an http server behind. Switching from `pkill` to `kill` seems to fix that.
- The test script runs on its own port to avoid conflicting with a dev server you might already have running.
- I killed the `npm run mocha-phantomjs` task, since it was so heavily tied to `scripts/test.sh` anyway.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/222)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,4 +9,4 @@ script: >
     npm run flow-check &&
     npm run lint &&
     npm run coverage &&
-    cat coverage/lcov.info | ./node_modules/.bin/coveralls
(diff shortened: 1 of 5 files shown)


armish:
looks great -- we were using pkill since that npm task was creating child processes that are not being killed immediately. Taking the mocha-phantomjs task out of the package description and putting it into the test sounds reasonable. 

Thanks for the fix!

:+1:

PR #223, opened 2015-07-08
Clean up GenomeTrack base pair rendering.

danvk:
Previously a `<rect>` and a `<text>` element were created for every base pair and one was hidden via CSS. Now I only create one. The trick is to add a wrapper element whose d3 data is the display mode. That way it will handle cleanup for us via `enter()`/`exit()`.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/223)


armish:
this one also looks good to me :)

PR #224, opened 2015-07-08
Show BAI fetch notification w/o index chunks

danvk:
Previously, when a BAI file didn't have index chunks, any requests for the BAM file header would block on downloading the entire BAI file (which is required to know how precisely large the BAM header is). This meant that there would be no status message visible while the BAI file was being loaded.

This change shows a message when that's the case, so that you always get an immediate notification. I tried to thread this notification through `bam.js` but had no luck.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/224)

### DIFF
--- a/src/main/BamDataSource.js
+++ b/src/main/BamDataSource.js
@@ -50,22 +50,31 @@ function createFromBamFile(remoteSource: BamFile): BamDataSource {
     }
   }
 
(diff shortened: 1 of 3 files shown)


armish:
LGTM -- I like the notification message (should help people trying to debug why things are not fast)

PR #227, opened 2015-07-08
Fix the flash of mismatches on initial load

danvk:
If the reference isn't available yet, then we shouldn't color mismatches based on it. This should eliminate the flash of color you occasionally see in the pileup tracks on the demo.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/227)

### DIFF
--- a/src/main/pileuputils.js
+++ b/src/main/pileuputils.js
@@ -77,7 +77,7 @@ function findMismatches(reference: string, seq: string, refPos: number, scores: 
     var pos = refPos + i,
         ref = reference.charAt(i),
         basePair = seq.charAt(i);
-    if (ref != basePair) {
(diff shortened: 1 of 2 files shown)


armish:
looks good!

In the long term, we might want to start handling base aliases properly. You can imagine a reference genome to consist of `N`s instead of `.` so that when you try to match any base with the reference, you won't get any mismatches.

danvk:
I'm distinguishing between an explicit `N` in the reference and `.` (meaning "not loaded yet"). If the reference is `N`, we might not want to display a mismatch either.

armish:
hmm, then the distinction sounds like a reasonable plan to me :+1:

PR #229, opened 2015-07-10
Display coverage information for BAM sources

armish:
This is a new track implementation that uses BAM format as input and plots a histogram-like graph that shows how many reads overlap with each nucleotide within the visible range. If the range changes (zoom in/out or panning), it adjusts the size of the bars and the information shown on the Y axis accordingly.

screen shot 2015-07-10 at 6 21 40 pm

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/229)

### DIFF
--- a/examples/playground.js
+++ b/examples/playground.js
@@ -24,6 +24,15 @@ var sources = [
     name: 'Genes'
   },
   {
+    viz: pileup.viz.coverage(),
+    data: pileup.formats.bam({
+      url: '/test-data/synth3.normal.17.7500000-7515000.bam',
+      indexUrl: '/test-data/synth3.normal.17.7500000-7515000.bam.bai'
+    }),
+    cssClass: 'normal-coverage',
+    name: 'Coverage'
+  },
+  {
     viz: pileup.viz.pileup(),
     data: pileup.formats.bam({
       url: '/test-data/synth3.normal.17.7500000-7515000.bam',
@@ -31,6 +40,15 @@ var sources = [
     }),
     cssClass: 'normal',
     name: 'Alignments'
+  },
+  {
+    viz: pileup.viz.coverage(),
+    data: pileup.formats.bam({
+      url: '/test-data/synth3.normal.17.7500000-7515000.bam',
+      indexUrl: '/test-data/synth3.normal.17.7500000-7515000.bam.bai'
+    }),
+    cssClass: 'tumor-coverage',
+    name: 'Coverage'
   },
   {
     viz: pileup.viz.pileup(),
--- a/examples/playground.js
+++ b/examples/playground.js
@@ -1,5 +1,11 @@
 var pileup = require('pileup');
 
+// We are going to use the same data source for multiple tracks
+var bamSource = pileup.formats.bam({
+  url: '/test-data/synth3.normal.17.7500000-7515000.bam',
+  indexUrl: '/test-data/synth3.normal.17.7500000-7515000.bam.bai'
+});
+
 var sources = [
   {
     viz: pileup.viz.genome(),
@@ -25,37 +31,25 @@ var sources = [
   },
   {
     viz: pileup.viz.coverage(),
-    data: pileup.formats.bam({
-      url: '/test-data/synth3.normal.17.7500000-7515000.bam',
-      indexUrl: '/test-data/synth3.normal.17.7500000-7515000.bam.bai'
-    }),
+    data: bamSource,
     cssClass: 'normal-coverage',
     name: 'Coverage'
   },
   {
     viz: pileup.viz.pileup(),
-    data: pileup.formats.bam({
-      url: '/test-data/synth3.normal.17.7500000-7515000.bam',
-      indexUrl: '/test-data/synth3.normal.17.7500000-7515000.bam.bai'
-    }),
+    data: bamSource,
     cssClass: 'normal',
     name: 'Alignments'
   },
   {
     viz: pileup.viz.coverage(),
-    data: pileup.formats.bam({
-      url: '/test-data/synth3.normal.17.7500000-7515000.bam',
-      indexUrl: '/test-data/synth3.normal.17.7500000-7515000.bam.bai'
-    }),
+    data: bamSource,
     cssClass: 'tumor-coverage',
     name: 'Coverage'
   },
   {
     viz: pileup.viz.pileup(),
-    data: pileup.formats.bam({
-      url: '/test-data/synth3.normal.17.7500000-7515000.bam',
-      indexUrl: '/test-data/synth3.normal.17.7500000-7515000.bam.bai'
-    }),
+    data: bamSource,
     cssClass: 'tumor',
     name: 'Alignments'
   }
--- a/examples/playground.js
+++ b/examples/playground.js
@@ -32,7 +32,7 @@ var sources = [
   {
     viz: pileup.viz.coverage(),
     data: bamSource,
-    cssClass: 'normal-coverage',
+    cssClass: 'normal',
     name: 'Coverage'
   },
   {
@@ -44,7 +44,7 @@ var sources = [
   {
     viz: pileup.viz.coverage(),
     data: bamSource,
-    cssClass: 'tumor-coverage',
+    cssClass: 'tumor',
     name: 'Coverage'
   },
   {
--- a/src/main/CoverageTrack.js
+++ b/src/main/CoverageTrack.js
@@ -224,16 +224,6 @@ class NonEmptyCoverageTrack extends React.Component {
       svg.append('g').attr('class', 'y-axis');
     } else {
       yAxisEl.call(yAxis);  // update the axis
-
-      /* aa
-      // Resize the background box according to the axis dimensions
-      var bbox = yAxisEl.node().getBBox();
-      svg.selectAll('rect.y-axis-background')
-        .attr('x', bbox.x)
-        .attr('y', bbox.y)
-        .attr('width', bbox.width * 1.2)  // %20 bigger box
-        .attr('height', bbox.height);
-      */
     }
   }
 }
--- a/src/main/CoverageTrack.js
+++ b/src/main/CoverageTrack.js
@@ -1,5 +1,5 @@
 /**
- * Pileup visualization of BAM sources.
+ * Coverage visualization of BAM sources.
  * @flow
  */
 'use strict';
@@ -46,6 +46,8 @@ class NonEmptyCoverageTrack extends React.Component {
   }
 
   render(): any {
+    // Fill the container up
+    // as we are going to use almost all the space available
     var containerStyles = {
       'height': '100%'
     };
@@ -80,7 +82,13 @@ class NonEmptyCoverageTrack extends React.Component {
     this.updateSize();
 
     var div = this.refs.container.getDOMNode();
-    d3.select(div).append('svg');
+    var svg = d3.select(div).append('svg');
+
+    // Below, we create the container group for our bars upfront
+    // as we want to overlay the axis on top of them
+    // therefore we have to make sure their container
+    // is defined first in the SVG
+    svg.append('g').attr('class', 'bin-group');
 
     this.props.source.on('newdata', () => {
       var range = this.props.range,
@@ -109,42 +117,85 @@ class NonEmptyCoverageTrack extends React.Component {
     // Hold off until height & width are known.
     if (width === 0) return;
 
-    svg
-      .attr('width', width)
-      .attr('height', height);
+    svg.attr('width', width).attr('height', height);
 
+    // Calculate summary statistics for coverage from the reads
     var binCounts = _.chain(this.state.reads)
+      // Extract the interval a read covers
       .map(read => read.getInterval().interval)
+      // Create arrays for nucleotides covered by a read
       .map(interval => _.range(interval.start, interval.stop+1))
+      // Merge all covered reads into a single array
       .flatten()
+      // Group observations by nucleotide position and count them
       .countBy()
+      // Convert the pos=>count object into an array for easier parsing
       .pairs()
+      // Convert the array into a custom object to be used for D3 plotting
       .map(c => ({key: c[0], count: c[1]}))
+      // Extract the value from the underscore chain
       .value();
 
-    var maxVal = _.chain(binCounts).map(b => b.count).max().value();
+    var maxCoverage = _.chain(binCounts)
+      .map(b => b.count)
+      .max()
+      .value();
+    // Now that we know the max coverage,
+    // we now want to create a visually appealing axis
+    // to make it easy to comprehend for us, humans
+    // Rule: if maxCoverage is smaller than 10X,
+    //  then show it as it is (we like small numbers);
+    //  otherwise, round the number to the nearest tenth
+    //  (we hate numbers like 61)
+    maxCoverage = maxCoverage < 10
+      ? maxCoverage
+      : Math.round(maxCoverage/10)*10;
+
+    // We are also going to add some padding in two ways:
+    //   1. leave some padding at the bottom for axis origin text (axisHeight)
+    //   2. do not create an axis with the range = maxCoverage (axisMax)
+    var axisMax = maxCoverage * 1.2;  // +20% of the original value
+    var axisHeight = height * .9;  // use 90% of the available space
     var yScale = d3.scale.linear()
-      .domain([0, maxVal])
-      .range([0, height]);
+      .domain([axisMax, 0])  // mind the inverted axis
+      .range([0, axisHeight]);
 
-    var histBars = svg.selectAll("rect.covbin").data(binCounts, d => d.key);
+    // Select the group we created first
+    var histBars = svg.select('g.bin-group').selectAll("rect.covbin")
+      .data(binCounts, d => d.key);
 
+    // D3 logic for our histogram bars
     histBars
       .enter()
       .append('rect')
       .attr('x', d => xScale(d.key))
-      .attr('y', d => yScale(maxVal-d.count))
+      .attr('y', d => yScale(d.count))
       .attr('width', d => xScale(d.key) - xScale(d.key-1))
-      .attr('height', d => yScale(d.count))
+      .attr('height', d => yScale(axisMax-d.count))
       .attr('class', 'covbin');
-
     histBars
       .attr('x', d => xScale(d.key))
-      .attr('y', d => yScale(maxVal-d.count))
+      .attr('y', d => yScale(d.count))
       .attr('width', d => xScale(d.key) - xScale(d.key-1))
-      .attr('height', d => yScale(d.count))
-
+      .attr('height', d => yScale(axisMax-d.count))
     histBars.exit().remove();
+
+    // Logic for our axis
+    var yAxis = d3.svg.axis()
+      .scale(yScale)
+      .orient('left')  // this is gonna be at the far right
+      .tickSize(5)  // Make our ticks much more visible
+      .outerTickSize(0)  // Remove the default range ticks (they are ugly)
+      .tickFormat(t => t + 'X')  // X -> times in coverage terminology
+      .tickValues([0, maxCoverage/2, maxCoverage]);  // show min, avg and max
+    var yAxisEl = svg.selectAll('g.y-axis');
+    if(yAxisEl.empty()) {  // no axis element yet
+      svg.append('g')
+        .attr('class', 'y-axis')
+        .attr('transform', 'translate(' + width + ', 0)');
+    }
+    yAxisEl.call(yAxis);  // update the axis
+
   }
 }
 
--- a/src/main/CoverageTrack.js
+++ b/src/main/CoverageTrack.js
@@ -71,34 +71,35 @@ class NonEmptyCoverageTrack extends React.Component {
     });
   }
 
+  /**
+   * Create a dummy label to see how much space the label occupies.
+   * Once calculated, save it into the state for later use in
+   * padding-related calculations.
+   */
+  calculateLabelSize(svg) {
+    // Create a dummy element that matches with the style:
+    //        ".coverage .y-axis g.tick text {...}"
+    var dummyAxis = svg.append('g').attr('class', 'y-axis');
+    var dummySize = dummyAxis.append('g').attr('class', 'tick')
+      .append('text')
+      .text("100X")
+      .node()
+      .getBBox();  // Measure its size
+
+    // Save the size information
+    this.setState({
+          labelSize: {height: dummySize.height, width: dummySize.width}
+    });
+    dummyAxis.remove();
+  }
+
   componentDidMount() {
     window.addEventListener('resize', () => this.updateSize());
     this.updateSize();
 
     var div = this.refs.container.getDOMNode();
     var svg = d3.select(div).append('svg');
-
-    /*
-      We are going to create a dummy label to see how big is our label
-      text will be. Once we get this information, we will save it in our state
-      and use it in plot padding calculations.
-    */
-    // Create a dummy element that matches with the style:
-    //        ".coverage .y-axis g.tick text {...}"
-    var dummyAxis = svg.append('g').attr('class', 'y-axis');
-    // Measure its size
-    var dummySize = dummyAxis.append('g').attr('class', 'tick')
-      .append('text')
-      .text("100X")
-      .node()
-      .getBBox();
-    // Save the size information
-    this.setState({
-          labelSize: {height: dummySize.height, width: dummySize.width}
-    });
-    // Remove the dummy element
-    dummyAxis.remove();
-    /* end of dummy label calculations */
+    this.calculateLabelSize(svg);
 
     // Below, we create the container group for our bars upfront as we want
     // to overlay the axis on top of them; therefore, we have to make sure
@@ -121,39 +122,35 @@ class NonEmptyCoverageTrack extends React.Component {
     }
   }
 
-  /*
-    The following extacts the summary statistics from the read data.
-    It might look ugly and you might have the temptation to convert
-    this into a functional form; but, please don't. We need these hacky
-    optimizations not to chug the browser with histogram data generation.
-  */
-  extractSummaryStatistics(reads: Array<SamRead>, range: GenomeRange) {
+  /**
+   * Extract summary statistics from the read data.
+   */
+  extractSummaryStatistics(reads: Array<SamRead>, range: ContigInterval<string>) {
     // Keep track of the start/stop points of our view
     var rstart = range.start - 1,  // this is to account for 0 vs 1-based pos.
         rstop = range.stop;
     // Create an array to keep track of all counts for each of the positions
-    var binCounts = new Array(rstop - rstart + 1);  // length = num of bases
-    binCounts = _.map(binCounts, () => 0);  // start w/ 0 counts for each base
-    _.chain(this.state.reads)  // Parse the read data
-        // Walk over the interval one base at a time and update the counts
-        .each(read => {
-            var interval = read.getInterval().interval,
-                istart = interval.start,
-                istop = interval.stop;
+    var binCounts = _.range(rstart, rstop + 1).map(() => 0);
 
-            for (var j = Math.max(istart, rstart);  // don't go beyond start
-                j <= Math.min(istop, rstop);  // don't go beyond stop
-                j++) {
-              binCounts[j - rstart] += 1;
-            }
-        });
-    // binCounts is a simple array now, so let's find the max val right away
+    // This is written in an imperative style (instead of with _.groupBy)
+    // as an optimization.
+    _.each(this.state.reads, read => {
+      var interval = read.getInterval().interval,
+          istart = interval.start,
+          istop = interval.stop;
+
+      for (var j = Math.max(istart, rstart);  // don't go beyond start
+          j <= Math.min(istop, rstop);  // don't go beyond stop
+          j++) {
+        binCounts[j - rstart] += 1;
+      }
+    });
     var maxCoverage = _.max(binCounts);
 
-    // This will convert the array into an array of objects with keys
-    // where the key will be the nucleotide location
+    // This will convert the array into an array of objects with positions
+    // where the position will be used a unique key
     binCounts = _.map(binCounts,
-      (val, idx) => ({key: rstart + idx + 1, count: val})
+      (val, idx) => ({position: rstart + idx + 1, count: val})
     );
 
     return {binCounts, maxCoverage};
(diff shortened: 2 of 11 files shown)


danvk:
Thanks for adding this much-needed feature, Arman. It looks great!

My one high-level comment is about performance. If I open the demo page in Chrome w/ the dev tools closed and drag back and forth to pan, I see the frame rate drop from ~12fps→2fps with this change. I have one theory about why (see my comment about  DOM measurements). It's not unexpected that this change would have a negative impact on performance, but I wouldn't expect it to be quite that dramatic. Some profiling might shed light on this. I'm happy to show you how to do that with the Chrome dev tools if you haven't played around with them before.

---

Review status: 0 of 6 files reviewed at latest revision, 23 unresolved discussions, all commit checks successful.

---

<sup>**[examples/playground.js, line 47 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju6xYlWN6r10XR4ZxIQ)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/examples/playground.js#L47)):</sup>
Just `tumor` should be fine—the `coverage` class should come from the visualization component.

---

<sup>**[src/main/CoverageTrack.js, line 2 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju6xmoHItgA5pC7fVrd)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L2)):</sup>
BAM→Alignment, given my upcoming GA4GH changes

---

<sup>**[src/main/CoverageTrack.js, line 68 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju6ySSYED7lgYBSHI8R)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L68)):</sup>
This snippet should go into some kind of utils module—it's shared across every single module (sorry, my fault for all the copy/paste code).

---

<sup>**[src/main/CoverageTrack.js, line 87 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju6z3Yg1_OUYHTYeyAg)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L87)):</sup>
nit: it looks like your comments are wrapping around 60 chars, rather than 80

---

<sup>**[src/main/CoverageTrack.js, line 123 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju6zhCl-uTLOJu2TeiH)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L123)):</sup>
Nice use of `_.chain` and great comments here. I'd suggest pulling this out into a separate function for more direct testing.

---

<sup>**[src/main/CoverageTrack.js, line 139 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju6zoLuzx8f3knKjus7)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L139)):</sup>
Isaac & my rule of thumb is that `_.chain` is usually only worthwhile if you're doing at least three or four operations (like the six you do to get `binCounts`), so

```javascript
var maxCoverage = _.max(binCounts.map(b => b.count));
```

up to you.

---

<sup>**[src/main/CoverageTrack.js, line 143 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju70mkm_I0v9uscmCnm)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L143)):</sup>
I'm not sure that the first three lines of comment here add much. Is there ever a situation where we don't want a visually appealing axis? :)

---

<sup>**[src/main/CoverageTrack.js, line 150 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju7-llqDG-7ED4mweIj)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L150)):</sup>
Can you use d3's [`nice`][1] method in place of these heuristics?

[1]: https://github.com/mbostock/d3/wiki/Quantitative-Scales#linear_nice

---

<sup>**[src/main/CoverageTrack.js, line 152 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju71C6R5Yi-uJsdKesk)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L152)):</sup>
we typically put spaces around operators

---

<sup>**[src/main/CoverageTrack.js, line 156 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju74ynKhRg0LgoOPCs8)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L156)):</sup>
I'm not sure I follow this.

---

<sup>**[src/main/CoverageTrack.js, line 158 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju7511V_F3YPm-J8hjG)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L158)):</sup>
Would it be better to make the range something like `[0, axisHeight - 6]`? The amount of space you need to reserve will change depending on the label font size, not the height of the track.

The ideal way to do this would be to create a fake label in `componentDidMount`, measure it's height and then do the layout precisely. It's actually not too hard to do this!

---

<sup>**[src/main/CoverageTrack.js, line 175 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju75eNojtuGKC5I_MBV)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L175)):</sup>
How about `coverage-bin`, or just `bin`?

---

<sup>**[src/main/CoverageTrack.js, line 187 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju76RB3rnmkcXKdTUEC)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L187)):</sup>
I was looking up what this did in the [D3 reference][1]. It looks like it's supposed to take one argument which is a two element array?

(one advantage of typescript over flow is that we'd be able to use their [d3 typings][2])

[1]: https://github.com/mbostock/d3/wiki/SVG-Axes#tickSize
[2]: https://github.com/borisyankov/DefinitelyTyped/blob/master/d3/d3.d.ts

---

<sup>**[src/main/CoverageTrack.js, line 189 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju75s-Oi1pWdNQ3Uzut)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L189)):</sup>
is capital `X` or lowercase `x` more typical in this situation?

---

<sup>**[src/main/CoverageTrack.js, line 190 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju76fIe7Q9UqzKL8sh7)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L190)):</sup>
nice! is this what IGV does, too?

---

<sup>**[src/main/CoverageTrack.js, line 192 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju76zet_81LPRkCm1m5)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L192)):</sup>
space after `if`

---

<sup>**[src/main/CoverageTrack.js, line 192 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju774PTUlYigBkXh470)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L192)):</sup>
Is this how axes are typically done in d3?

---

<sup>**[src/main/CoverageTrack.js, line 200 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju77aIgnQrLQS4hvZfC)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L200)):</sup>
Measuring the size of DOM elements on every frame can be catastrophic for performance, since if forces the browser to flush out changes to the DOM when it could otherwise batch updates.

One alternative to the translucent background is to put a white stroke around the text, so that it shows up more clearly against varying backgrounds. If I add [this CSS][1]:

```
  stroke: whitesmoke;
  paint-order: stroke;
```

and kill the background element, I get this:

screen shot 2015-07-13 at 12 04 17 pm

We should get Jackie's input on this.

[1]: http://stackoverflow.com/a/26498819/388951

---

<sup>**[src/test/CoverageTrack-test.js, line 2 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju7AaB0t9wtKQh9nv1I)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/test/CoverageTrack-test.js#L2)):</sup>
Copy/paste comment? I don't think this describes this test suite.

---

<sup>**[src/test/CoverageTrack-test.js, line 45 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju7AieAxmfOnlRhGKwY)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/test/CoverageTrack-test.js#L45)):</sup>
two character indent

---

<sup>**[src/test/CoverageTrack-test.js, line 53 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju7AoTSZnLqrPudfd2E)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/test/CoverageTrack-test.js#L53)):</sup>
Move into `beforeEach`?

---

<sup>**[src/test/CoverageTrack-test.js, line 90 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju7B4Gf2Eqv7knwzEad)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/test/CoverageTrack-test.js#L90)):</sup>
any idea why this is? `if` statements like this in tests are generally a bad idea. I'd either track down the issue or kill the assertion.

---

<sup>**[style/pileup.css, line 171 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju7BJDx5VkIP1RjHWhB)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/style/pileup.css#L171)):</sup>
Any reason to go 70% opaque gray vs. `fill: rgb(0.5, 0.5, 0.5)`, or whatever it comes out to? There shouldn't be anything underneath the bars and rendering opaque elements might be faster.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/229)

armish:
thanks for the comments, Dan! 

Let me investigate the performance issues with this implementation as you suggested. It has been almost a year since I last used the profiling feature within the dev tools; so I might need to blow some dust off of my ancient knowledge, but I will let you know if I get stuck with something :)

---







Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/229)

armish:
Review status: 0 of 9 files reviewed at latest revision, 12 unresolved discussions, some commit checks pending.

---

<sup>**[src/main/CoverageTrack.js, line 68 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju6ySSYED7lgYBSHI8R)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L68)):</sup>
this is done.

---

<sup>**[src/main/CoverageTrack.js, line 143 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju70mkm_I0v9uscmCnm)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L143)):</sup>
true that :)

---

<sup>**[src/main/CoverageTrack.js, line 150 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju7-llqDG-7ED4mweIj)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L150)):</sup>
I can definitely replace the logic for calculating the`axisMax` with a `nice()` call on the axis, but I then need to extract the value back from the axis to properly position/extend the bar graphs. Let me make the change and explain this in the comments for future reference (in case somebody else dives into the code and tries to make sense out of it)

The code for rounding the `maxCoverage` to the nearest tenth is necessary as we have a hard threshold (10X) for that one. 

---

<sup>**[src/main/CoverageTrack.js, line 156 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju74ynKhRg0LgoOPCs8)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L156)):</sup>
this is obsolete now, but sorry about the fuzzy language there

---

<sup>**[src/main/CoverageTrack.js, line 158 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju7511V_F3YPm-J8hjG)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L158)):</sup>
implemented your suggestion -- it was really not that hard :)

---

<sup>**[src/main/CoverageTrack.js, line 175 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju75eNojtuGKC5I_MBV)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L175)):</sup>
went w/ `bin`.

---

<sup>**[src/main/CoverageTrack.js, line 187 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju76RB3rnmkcXKdTUEC)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L187)):</sup>
ah, I meant to set the size of the innerTickSize -- fixed this.

---

<sup>**[src/main/CoverageTrack.js, line 189 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju75s-Oi1pWdNQ3Uzut)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L189)):</sup>
Both is OK, but I see capital `X` being more widely used in papers.

---

<sup>**[src/main/CoverageTrack.js, line 190 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju76fIe7Q9UqzKL8sh7)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L190)):</sup>
it shows a simpler axis:

Imgur

I think ours is much better ;)

---

<sup>**[src/main/CoverageTrack.js, line 192 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju774PTUlYigBkXh470)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L192)):</sup>
Yeah, axises are weird in the sense that they don't have an **enter** method; so people usually add the axis while initializing the plot and update it with the scale in a separate function. Instead of updating the label in a different function and to not increase the complexity of the code, I decided to make use of a conditional statement here.

---

<sup>**[src/main/CoverageTrack.js, line 200 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju77aIgnQrLQS4hvZfC)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/main/CoverageTrack.js#L200)):</sup>
cool trick -- thanks for pointing this out! Removed the background el and implemented this instead.

---

<sup>**[src/test/CoverageTrack-test.js, line 90 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju7B4Gf2Eqv7knwzEad)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/test/CoverageTrack-test.js#L90)):</sup>
after some more debugging, I found that phantomJS was evaluating the page faster and catching a state where no coverage bins were loaded up. I fixed this by modifying the promise so that it always waits until the coverage tracks is fully loaded.

So this conditional statement is no more.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/229)

danvk:
This is looking much better! What's the performance difference looking like now?

---

Review status: 0 of 9 files reviewed at latest revision, 18 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/CoverageTrack.js, line 81 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-JuD2tX_R8emIQfRTXiE)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0d8680b7038c3ada45e08f6c66762b44cca55fd2/src/main/CoverageTrack.js#L81)):</sup>
This is a self-contained task, so it would make sense to move this into its own method.

---

<sup>**[src/main/CoverageTrack.js, line 85 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-JuD34WFGsHqnBKNxl9g)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0d8680b7038c3ada45e08f6c66762b44cca55fd2/src/main/CoverageTrack.js#L85)):</sup>
We tend to only use `/* .. */`-style comments for jsdoc (here and elsewhere)

---

<sup>**[src/main/CoverageTrack.js, line 124 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-JuD4MwpjGFUH6T7gp0X)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0d8680b7038c3ada45e08f6c66762b44cca55fd2/src/main/CoverageTrack.js#L124)):</sup>
I'd split this into an API comment and an implementation comment. The former can use jsdoc style and go outside the method:

```javascript
/**
 * Extract summary statistics from the read data.
 */
```

The other relates to the implementation and should be in an inline comment inside the method:

```javascript
// This is written in an imperative style (instead of with _.groupBy) as an optimization.
```

---

<sup>**[src/main/CoverageTrack.js, line 126 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-JuD4wo6CliuwyXPu9zJ)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0d8680b7038c3ada45e08f6c66762b44cca55fd2/src/main/CoverageTrack.js#L126)):</sup>
This is a bit too chatty — I'd suggest stating what you mean more directly. See my suggestion above.

---

<sup>**[src/main/CoverageTrack.js, line 130 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-JuD3qNPSC9kv35Stu2P)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0d8680b7038c3ada45e08f6c66762b44cca55fd2/src/main/CoverageTrack.js#L130)):</sup>
Prefer `ContigInterval<string>` to `GenomeRange`. I'd like to deprecate the latter.

---

<sup>**[src/main/CoverageTrack.js, line 134 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-JuD5cFyr-oesf52Bx6r)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0d8680b7038c3ada45e08f6c66762b44cca55fd2/src/main/CoverageTrack.js#L134)):</sup>
This code is a bit over-commented. Most of these statements are fairly self-explanatory. Inline comments are most useful when there's some sort of subtlety that might not be apparent from reading the code. (your comment about 0 vs 1-based above is a great example)

---

<sup>**[src/main/CoverageTrack.js, line 135 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-JuD5Rt7SGiERwgK_bSs)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0d8680b7038c3ada45e08f6c66762b44cca55fd2/src/main/CoverageTrack.js#L135)):</sup>
You could also express this as:

```javascript
var binCounts = _.range(rstart, rstop + 1).map(() => 0);
```

if you prefer.

---

<sup>**[src/main/CoverageTrack.js, line 137 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-JuD5x6gjlCbTOt_lLQn)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0d8680b7038c3ada45e08f6c66762b44cca55fd2/src/main/CoverageTrack.js#L137)):</sup>
`_.chain` is unnecessary here. Just do `_.each(this.states.reads, read => { ... })` or `this.states.reads.forEach(read => { ... })`.

---

<sup>**[src/main/CoverageTrack.js, line 156 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-JuD6LLBkC80vo_jdT2l)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0d8680b7038c3ada45e08f6c66762b44cca55fd2/src/main/CoverageTrack.js#L156)):</sup>
`pos` would be a more precise name than `key` here.

---

<sup>**[src/main/CoverageTrack.js, line 189 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-JuD6zouedg7uRDzQHeN)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0d8680b7038c3ada45e08f6c66762b44cca55fd2/src/main/CoverageTrack.js#L189)):</sup>
The parens are superfluous on one parameter arrow functions. This works just as well:

```javascript
var calcBarHeight = d => Math.max(0, yScale(axisMax - d.count));
```

---

<sup>**[src/main/CoverageTrack.js, line 190 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-JuD7PrKHPEAI0wBXlkd)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0d8680b7038c3ada45e08f6c66762b44cca55fd2/src/main/CoverageTrack.js#L190)):</sup>
Pos → Y? Pos could be either coordinate.

---

<sup>**[src/main/CoverageTrack.js, line 190 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-JuD7VEjz5VVLUJeP4sa)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0d8680b7038c3ada45e08f6c66762b44cca55fd2/src/main/CoverageTrack.js#L190)):</sup>
Can you incorporate the top/bottom inversion directly into the scale.

---

<sup>**[src/main/CoverageTrack.js, line 191 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-JuD7LYPP6mS-vPXDRMW)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0d8680b7038c3ada45e08f6c66762b44cca55fd2/src/main/CoverageTrack.js#L191)):</sup>
Shouldn't this just be a constant?

---

<sup>**[src/main/CoverageTrack.js, line 200 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-JuD7hQ_1fj880gcgxXQ)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0d8680b7038c3ada45e08f6c66762b44cca55fd2/src/main/CoverageTrack.js#L200)):</sup>
You can drop these four `attr` calls—it should be sufficient to have them in the update selection.

---

<sup>**[src/main/GenomeTrack.js, line 119 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-JuD7vHP86SMnm2QT7Jv)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0d8680b7038c3ada45e08f6c66762b44cca55fd2/src/main/GenomeTrack.js#L119)):</sup>
It looks like this measures the width and then throws it away?

---

<sup>**[src/main/utils.js, line 169 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-JuD83-zF9lKahBK6fbz)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0d8680b7038c3ada45e08f6c66762b44cca55fd2/src/main/utils.js#L169)):</sup>
How about putting this in a `d3utils` module instead? It looks like the rest of this module is DOM-free.

---

<sup>**[src/main/utils.js, line 170 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-JuD8GcRZTm_3jdkCTMv)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0d8680b7038c3ada45e08f6c66762b44cca55fd2/src/main/utils.js#L170)):</sup>
space before `(`

---

<sup>**[src/test/CoverageTrack-test.js, line 53 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju7AoTSZnLqrPudfd2E)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/test/CoverageTrack-test.js#L53)):</sup>
I think it would be clearer to move the contents of this function into `beforeEach` and kill `testSetup`. There's no coherent distinction being drawn between the two right now: code just flows from one into the other.

---

<sup>**[src/test/CoverageTrack-test.js, line 90 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-Ju7B4Gf2Eqv7knwzEad)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f2914c6b227721951b0d3858daba389e0490a1eb/src/test/CoverageTrack-test.js#L90)):</sup>
Nicely done!

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/229)

armish:
It is much better: the FPS doesn't drop sharply anymore when you drag the reference track, but you can still feel that the browser is struggling to catch up with two pileup and two coverage tracks :( Interacting with a single pileup together with a single coverage track is smooth enough, though.

---

Review status: 0 of 9 files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/CoverageTrack.js, line 135 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-JuD5Rt7SGiERwgK_bSs)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0d8680b7038c3ada45e08f6c66762b44cca55fd2/src/main/CoverageTrack.js#L135)):</sup>
yeah, I like this one much better.

---

<sup>**[src/main/CoverageTrack.js, line 190 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-JuD7VEjz5VVLUJeP4sa)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0d8680b7038c3ada45e08f6c66762b44cca55fd2/src/main/CoverageTrack.js#L190)):</sup>
Sure, it is possible; but if we do that, we then have to hack the axis and interrupt with its labels/paths. I like having the axis as it is and dealing with the inverted rect calculations.

---

<sup>**[src/main/CoverageTrack.js, line 191 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-JuD7LYPP6mS-vPXDRMW)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0d8680b7038c3ada45e08f6c66762b44cca55fd2/src/main/CoverageTrack.js#L191)):</sup>
we can do that; but if we ever change our scaling from linear to another form of transformation (like a step function), this calculation will still hold nicely. It is somewhat redundant right now, but can save us from some headache in the future.

---

<sup>**[src/main/GenomeTrack.js, line 119 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-JuD7vHP86SMnm2QT7Jv)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0d8680b7038c3ada45e08f6c66762b44cca55fd2/src/main/GenomeTrack.js#L119)):</sup>
argh! Really good catch, sorry about missing this one :\

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/229)

danvk:
You've got some Flow errors that are breaking the Travis build. Other than that (and my one comment on `GenomeTrack`), this is good to merge!

---

Review status: 0 of 10 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

<sup>**[src/main/GenomeTrack.js, line 120 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-JuD7vHP86SMnm2QT7Jv)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0d8680b7038c3ada45e08f6c66762b44cca55fd2/src/main/GenomeTrack.js#L120)):</sup>
We should actually drop the measurement from `getScale`, since it gets called on every frame. We could add a resize listener to track this more sanely.

(sorry, this isn't really related to your change—I'm happy to do it in a follow-up if you prefer.)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/229)

armish:
Thanks Dan! I fixed the issue with the `GenomeTrack`. Also had to convert the type definitions from `ContigInterval<string>` back to `GenomeRange` for now as they turn out to be not inter-changable right away. I will create an issue to deprecate the use of `GenomeRange` from the whole code base and will get that done as a separate task.

---

Review status: 0 of 10 files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

<sup>**[src/main/GenomeTrack.js, line 120 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/229#-JuD7vHP86SMnm2QT7Jv)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0d8680b7038c3ada45e08f6c66762b44cca55fd2/src/main/GenomeTrack.js#L120)):</sup>
turns out -- the logic was already there and this was redundant. Dropped it while I am at it ;)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/229)

danvk:
Looks great, thanks for implementing this much-needed feature!

PR #232, opened 2015-07-15
Optimizations for alignment tracks

danvk:
This makes `BamDataSource`'s `newdata` event fire only when new data becomes available (previously it fired when a cached region was requested, too). This should significantly reduce re-computation in `PileupTrack`.

I also reworked the data flow in `CoverageTrack` so that it only computes summary statistics when `newdata` fires.

Performance numbers while panning on the playground demo:
Before: 4-5fps
After: 9-10fps

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/232)

### DIFF
--- a/src/main/BamDataSource.js
+++ b/src/main/BamDataSource.js
@@ -86,6 +86,7 @@ function createFromBamFile(remoteSource: BamFile): AlignmentDataSource {
           coveredRanges = ContigInterval.coalesce(coveredRanges);
           reads.forEach(read => addRead(read));
           o.trigger('networkdone');
+          o.trigger('newdata', range);
         });
     });
   }
@@ -102,9 +103,7 @@ function createFromBamFile(remoteSource: BamFile): AlignmentDataSource {
 
   var o = {
     rangeChanged: function(newRange: GenomeRange) {
-      fetch(newRange)
-          .then(() => o.trigger('newdata', newRange))
-          .done();
+      fetch(newRange).done();
     },
     getAlignmentsInRange,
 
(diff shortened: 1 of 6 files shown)


armish:
looks great! Thanks for this optimization, Dan!

---

Reviewed 3 of 3 files at r1.

---

<sup>**[src/main/CoverageTrack.js, line 151 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/232#-JuHFhQjImliIoIf7-CA)** ([raw file](https://github.com/hammerlab/pileup.js/blob/471c7243aef798819d4c5622c49972a979e13dc8/src/main/CoverageTrack.js#L151)):</sup>
Do we really need to sort this array? Is `.filter` smart enough to take this sorting into account?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/232)

danvk:
I noticed a bug where this made `PileupTrack` not render some alignments that it knew about. I'll try to track that down.

---

---

<sup>**[src/main/CoverageTrack.js, line 151 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/232#-JuHFhQjImliIoIf7-CA)** ([raw file](https://github.com/hammerlab/pileup.js/blob/471c7243aef798819d4c5622c49972a979e13dc8/src/main/CoverageTrack.js#L151)):</sup>
We don't, but in theory we could use binary search in the future. More practically, this means that the bars are arranged in order in the DOM, which I think is a nice property.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/232)

danvk:
@armish mind taking another look? I fixed the issue with missing alignments.

armish:
the changes look good :+1:

---

Reviewed 4 of 4 files at r2.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/232)

PR #236, opened 2015-07-20
Fix the version for mocha-phantomjs dependency

armish:
The newer version of the `mocha-phantomjs` (v3.6.0) requires an older version of phantomjs and this create a dependency conflict, causing [our build to fail](https://travis-ci.org/hammerlab/pileup.js/builds/71798433). This change fixes the version of our `mocha-phantomjs` dependency to resolve that issue.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/236)


danvk:
Feel free to merge if the tests pass.

PR #237, opened 2015-07-20
Consolidate duplicated code across Visualization tracks

danvk:
This eliminates two sources of duplicated code:

- The empty / non-empty track dance that each component had to do. This is now done by a new component, `VisualizationWrapper`.
- Caching the track's size. The last-measured size is passed down via `props.{width,height}` now.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/237)

### DIFF
--- a/src/main/CoverageTrack.js
+++ b/src/main/CoverageTrack.js
@@ -15,27 +15,35 @@ var React = require('./react-shim'),
     _ = require("underscore"),
     ContigInterval = require('./ContigInterval');
 
-var CoverageTrack = React.createClass({
-  displayName: 'coverage',
-  propTypes: {
-    range: types.GenomeRange,
-    onRangeChange: React.PropTypes.func.isRequired,
-    source: React.PropTypes.object.isRequired,
-    referenceSource: React.PropTypes.object.isRequired
-  },
-  render: function(): any {
-    var range = this.props.range;
-    if (!range) {
-      return <EmptyTrack />;
+
+/**
+ * Extract summary statistics from the read data.
+ */
+function extractSummaryStatistics(reads: Array<SamRead>, contig: string) {
+  var binCounts = {};
+
+  // This is written in an imperative style (instead of with _.groupBy)
+  // as an optimization.
+  _.each(reads, read => {
+    var interval = read.getInterval();
+
+    var start = interval.start(),
+        stop = interval.stop();
+    for (var j = start; j <= stop; j++) {
+      binCounts[j] = (binCounts[j] || 0) + 1;
     }
+  });
+  var maxCoverage = _.max(binCounts);
 
-    return <NonEmptyCoverageTrack {...this.props} />;
-  }
-});
+  binCounts = _.map(binCounts, (count, position) => ({position: Number(position), count}));
+  binCounts = _.sortBy(binCounts, ({position, count}) => position);
+
+  return {binCounts, maxCoverage};
+}
 
 
-class NonEmptyCoverageTrack extends React.Component {
-  constructor(props) {
+class CoverageTrack extends React.Component {
+  constructor(props: Object) {
     super(props);
     this.state = {
       width: 0,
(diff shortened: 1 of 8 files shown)


armish:
Really cool change, Dan! Loved the way it made all the tracks much easier to read/understand.

I added a few comments/questions just to make sure I get them right, but other than that it looks good to go!

---

Reviewed 8 of 8 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/CoverageTrack.js, line 22 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/237#-Jugu1NctyeaqzjfilNS)** ([raw file](https://github.com/hammerlab/pileup.js/blob/9c691a4a92d436304498637b0de74cf67dbcf700/src/main/CoverageTrack.js#L22)):</sup>
Is there a specific reason as to why we  make these utility functions external to the React class? (you did a similar thing with the other tracks as well, I guess).

---

<sup>**[src/main/GeneTrack.js, line 219 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/237#-JugvBARKsG8yoc3fsMz)** ([raw file](https://github.com/hammerlab/pileup.js/blob/4f26993643fb4314568340ddb81b762e5868225d/src/main/GeneTrack.js#L219)):</sup>
With this change, we lose the ability to define custom `EmptyTrack` messages for each track, right? Were they just placeholders (and you removed them intentionality) or are they gone by mistake?

---

<sup>**[src/main/PileupTrack.js, line 400 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/237#-JugueawhHaKJO1h4NYQ)** ([raw file](https://github.com/hammerlab/pileup.js/blob/9c691a4a92d436304498637b0de74cf67dbcf700/src/main/PileupTrack.js#L400)):</sup>
All the other tracks (expect `CoverageTrack` since it is a duplicate of this track) defines the `displayName` within the scope of the React class definition. Haven't realized this before -- should we make this consistent across all tracks?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/237)

danvk:
Review status: all files reviewed at latest revision, 3 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/CoverageTrack.js, line 22 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/237#-Jugu1NctyeaqzjfilNS)** ([raw file](https://github.com/hammerlab/pileup.js/blob/9c691a4a92d436304498637b0de74cf67dbcf700/src/main/CoverageTrack.js#L22)):</sup>
It's basically a way to make them private to the module. Now that I'm exporting `CoverageTrack`, Flow wants me to type annotate all of its methods. I'd prefer not to document the precise return type of this method (Flow can figure it out). Moving it out of the class removes the need to explicitly document the type.

---

<sup>**[src/main/PileupTrack.js, line 400 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/237#-JugueawhHaKJO1h4NYQ)** ([raw file](https://github.com/hammerlab/pileup.js/blob/9c691a4a92d436304498637b0de74cf67dbcf700/src/main/PileupTrack.js#L400)):</sup>
It's because we're using the old-style `React.createClass` syntax for some tracks and the new-style ES6 class syntax for others. We should probably switch to the ES6 syntax everywhere.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/237)

danvk:

---

<sup>**[src/main/GeneTrack.js, line 219 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/237#-JugvBARKsG8yoc3fsMz)** ([raw file](https://github.com/hammerlab/pileup.js/blob/4f26993643fb4314568340ddb81b762e5868225d/src/main/GeneTrack.js#L219)):</sup>
They weren't really accurate before -- the message was displayed when the reference source hadn't loaded yet, not when you needed to zoom in. We can restore them (via a property on the visualization) if need be.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/237)

PR #238, opened 2015-07-20
Reduce over-firing of newdata event for BAM files

danvk:
This change marks an interval as "covered" immediately after it's requested. This has the effect of preventing duplicate requests.

For example, if you were panning to the right, `BamDataSource` would repeatedly fetch the same interval until it got data back (and marked it as covered). Now it just fetches it once.

This is a pretty nice win. With all four pileup+coverage tracks, panning is noticeably smoother.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/238)


armish:
Cool! LGTM.

armish:
Another `TODO` bites the dust ;)

LGTM, feel free to merge this one in.

PR #251, opened 2015-09-16
Show paired reads in PileupTrack.

### LINKED ISSUE #108: pileup cache should be tied to PileupTrack + contig
Currently it's a [global variable][1] storing `Interval`s, but it should be tied to the particular PileupTrack (in case there's more than one, e.g. tumor+normal) and it should be tied to the current contig as well.

[1]: https://github.com/hammerlab/pileup.js/blob/4fcff3ab868e5319a3414f6c0eb174ea7830f8c1/src/PileupTrack.js#L65

### LINKED ISSUE #110: Show paired reads

danvk:
This factors out a PileupCache class, which manages the data being visualized by PileupTrack.

Here's what it looks like:
pairs in pileup

Remaining work, probably for follow-up PRs:
- Make "view as pairs" optional (includes some design & API work, too)
- Performance optimizations (there seems to be a bit of a regression)
- Pack the reads more intelligently (there's more wasted space now)

Fixes #110 
Fixes #108

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/251)

### DIFF
--- a/examples/playground.js
+++ b/examples/playground.js
@@ -64,6 +64,6 @@ var sources = [
 ];
 
 var p = pileup.create(yourDiv, {
-  range: {contig: 'chr17', start: 7512384, stop: 7512544},
+  range: {contig: 'chr17', start: 7512284, stop: 7512644},
   tracks: sources
 });
--- a/lib/underscore.js
+++ b/lib/underscore.js
@@ -1,6 +1,7 @@
 // type definitions for (some of) underscore
 
 declare module "underscore" {
+  declare function find<T>(list: T[], predicate: (val: T)=>boolean): ?T;
   declare function findWhere<T>(list: Array<T>, properties: {}): ?T;
   declare function clone<T>(obj: T): T;
 
@@ -38,5 +39,8 @@ declare module "underscore" {
 
   declare function min<T>(a: Array<T>|{[key:any]: T}): T;
   declare function max<T>(a: Array<T>|{[key:any]: T}): T;
+
+  declare function values<T>(o: {[key: any]: T}): T[];
+  declare function flatten(a: Array<any>): Array<any>;
 }
 
--- a/src/main/PileupCache.js
+++ b/src/main/PileupCache.js
@@ -1,0 +1,183 @@
+/**
+ * Data management for PileupTrack.
+ *
+ * This class groups paired reads and piles them up.
+ *
+ * @flow
+ */
+'use strict';
+
+import type {Strand, Alignment, AlignmentDataSource} from './Alignment';
+import type {TwoBitSource} from './TwoBitDataSource';
+import type {BasePair} from './pileuputils';
+
+var _ = require('underscore'),
+    ContigInterval = require('./ContigInterval'),
+    Interval = require('./Interval'),
+    {addToPileup, getOpInfo, CigarOp} = require('./pileuputils'),
+    utils = require('./utils');
+
+// This bundles everything intrinsic to the alignment that we need to display
+// it, i.e. everything not dependend on scale/viewport.
+export type VisualAlignment = {
+  read: Alignment;
+  strand: Strand;
+  refLength: number;  // span on the reference (accounting for indels)
+  mismatches: Array<BasePair>;
+};
+
+// This is typically a read pair, but may be a single read in some situations.
+type VisualGroup = {
+  key: string;
+  row: number;  // pileup row.
+  span: ContigInterval<string>;  // tip-to-tip span for the read group
+  insert: ?Interval;  // interval for the connector, if applicable.
+  alignments: VisualAlignment[];
+};
+
+// read name would make a good key, but paired reads from different contigs
+// shouldn't be grouped visually. Hence we use read name + contig.
+function groupKey(read: Alignment): string {
+  return `${read.getName()}:${read.ref}`;
+}
+
+// This class provides data management for the visualization, grouping paired
+// reads and managing the pileup.
+class PileupCache {
+  // maps groupKey to VisualGroup
+  groups: {[key: string]: VisualGroup};
+  refToPileup: {[key: string]: Array<Interval[]>};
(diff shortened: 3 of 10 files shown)


armish:
quickly skimmed through, but it looks good to me @danvk! 

---

Reviewed 11 of 11 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

<sup>**[src/main/GA4GHAlignment.js, line 8 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/251#-JzNCyzO-hQal37z1kQt)** ([raw file](https://github.com/hammerlab/pileup.js/blob/d958869e890b1e027baaacfa09b13b0d450d0cfe/src/main/GA4GHAlignment.js#L8)):</sup>
left-over?

---

<sup>**[src/main/PileupCache.js, line 76 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/251#-JzNDmOGPxxzbwe-XqhJ)** ([raw file](https://github.com/hammerlab/pileup.js/blob/d958869e890b1e027baaacfa09b13b0d450d0cfe/src/main/PileupCache.js#L76)):</sup>
Do groups ever have more than 2 elements? If so, do you think it is worth converting this into a hashmap-based comparison instead?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/251)

danvk:
Back to you, @armish!

---

Review status: 8 of 10 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

<sup>**[src/main/PileupCache.js, line 76 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/251#-JzNDmOGPxxzbwe-XqhJ)** ([raw file](https://github.com/hammerlab/pileup.js/blob/d958869e890b1e027baaacfa09b13b0d450d0cfe/src/main/PileupCache.js#L76)):</sup>
It's possible, but it's quite unusual (it would have to be a chimeric alignment). In the common case there will be 0, 1 or 2 alignments in the read group when `addAlignment` is called, in which case `_.find`'s linear search is the best approach.

---

<sup>**[src/main/GA4GHAlignment.js, line 8 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/251#-JzNCyzO-hQal37z1kQt)** ([raw file](https://github.com/hammerlab/pileup.js/blob/d958869e890b1e027baaacfa09b13b0d450d0cfe/src/main/GA4GHAlignment.js#L8)):</sup>
Done.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/251)

danvk:
I experimented with the performance a bit. The frame rates while panning the default playground view drop from ~14-15fps → ~12-13fps, i.e. a 20% decrease. It's possible that this is just because there's more stuff on the screen.

armish:
thanks, @danvk. I think the drop in the performance is still OK given the informative context this new feature brings to pileup.js. Have you tested this with non-paired reads? I wonder what the performance looks with those (to be addressed by your optimization PR in near future).

:+1:

---

Reviewed 2 of 3 files at r2.

---

<sup>**[src/main/PileupCache.js, line 76 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/251#-JzNDmOGPxxzbwe-XqhJ)** ([raw file](https://github.com/hammerlab/pileup.js/blob/d958869e890b1e027baaacfa09b13b0d450d0cfe/src/main/PileupCache.js#L76)):</sup>
yeah, that makes sense. Thanks for that clarifying that for me.

---

<sup>**[src/main/PileupTrack.js, line 325 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/251#-JzR3Nejmrx6hGmDkUbo)** ([raw file](https://github.com/hammerlab/pileup.js/blob/d958869e890b1e027baaacfa09b13b0d450d0cfe/src/main/PileupTrack.js#L325)):</sup>
just curious: why did the earlier piece of code not work? I would have thought it would work :)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/251)

danvk:
I haven't specifically tested it on unpaired reads. Any suggestions on a good source for them? i.e. a BAM file with lots of unpaired reads or a way to produce them?

---

---

<sup>**[src/main/PileupTrack.js, line 325 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/251#-JzR3Nejmrx6hGmDkUbo)** ([raw file](https://github.com/hammerlab/pileup.js/blob/d958869e890b1e027baaacfa09b13b0d450d0cfe/src/main/PileupTrack.js#L325)):</sup>
Yeah, me too.

It looks like [`attr`][1] can be passed:

- An attribute name and value
- An attribute name and a function which returns a value
- An object mapping attribute names to values

but it can't be passed a function which returns an object mapping attribute names to values. If you look at d3 issue 2109, they're going to fix this in d3 4.0.

[1]: https://github.com/mbostock/d3/wiki/Selections#attr

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/251)

armish:
Not off the top of m head. I always wanted to create those sample bam/2bit/genes files for easier testing and killing our dependency to external resources, but never was able to get to it. I will create an issue for this and look into it later on.

---

---

<sup>**[src/main/PileupTrack.js, line 325 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/251#-JzR3Nejmrx6hGmDkUbo)** ([raw file](https://github.com/hammerlab/pileup.js/blob/d958869e890b1e027baaacfa09b13b0d450d0cfe/src/main/PileupTrack.js#L325)):</sup>
:+1:

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/251)

Issue #255, opened 2015-09-18
D3/SVG → canvas


danvk:
I experimented with switching from SVG to Canvas in `PileupTrack`. It:

1. is simpler to code against than D3
2. is dramatically faster (at least 10x)
3. works better with Flow (to which d3 is opaque)

I'm clearing the entire canvas and redrawing it on every frame and _still_ getting 10x performance, so there are even more wins to be had if we need them.

Visualizations to port:

- [x] PileupTrack
- [x] CoverageTrack
- [x] GeneTrack
- [x] GenomeTrack
- [x] LocationTrack
- [x] ScaleTrack
- [ ] VariantTrack

Other cleanup:

- [x] Add back support for strandedness in GeneTrack

The components don't interact, so these can be done one at a time. We can also leave some of the less performance-intensive tracks (e.g. `VariantTrack`) as SVG if we want.

We lose the ability to color/style tracks using CSS when we do this. I think that's probably worth it.

PR #257, opened 2015-09-18
Migrate PileupTrack from SVG→canvas

danvk:
See #255 

When I remove all tracks but a single `PileupTrack`, I see panning performance at ~700bp scale go from ~20fps→60fps. It's better than that, though. The SVG version sometimes goes down to around 6fps while panning, whereas the canvas version never drops below 30fps.

Remaining bits to implement:
- [x] Make mismatched base pairs the correct font/size
- [x] Make mismatches turn into blocks of color
- [x] Add back click handler support
- [x] Update the tests to understand canvas

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/257)

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "chai": "^2.0.0",
     "coveralls": "2.10.x",
     "es5-shim": "^4.1.0",
-    "flow-bin": "^0.15.0",
+    "flow-bin": "^0.16.0",
     "http-server": "^0.8.0",
     "istanbul": "^0.3.17",
     "jscoverage": "^0.5.9",
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-noflow=$(git ls-files | egrep '^(src|test).*\.js$' | grep -v '/canvas.js' | xargs grep --files-without-match '@flow')
+noflow=$(git ls-files | egrep '^(src|test).*\.js$' | grep -v '/data-canvas.js' | xargs grep --files-without-match '@flow')
 if [ -n "$noflow" ]; then
   echo 'These files are missing @flow annotations:'
   echo "$noflow"
--- a/scripts/lint.sh
+++ b/scripts/lint.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-noflow=$(git ls-files | egrep '^(src|test).*\.js$' | xargs grep --files-without-match '@flow')
+noflow=$(git ls-files | egrep '^(src|test).*\.js$' | grep -v '/canvas.js' | xargs grep --files-without-match '@flow')
 if [ -n "$noflow" ]; then
   echo 'These files are missing @flow annotations:'
   echo "$noflow"
--- a/src/main/Alignment.js
+++ b/src/main/Alignment.js
@@ -7,10 +7,12 @@ import type * as ContigInterval from './ContigInterval';
 
 // "CIGAR" operations express how a sequence aligns to the reference: does it
 // have insertions? deletions? For more background, see the SAM/BAM paper.
+export type CigarSymbol = 'M'|'I'|'D'|'N'|'S'|'H'|'P'|'='|'X';
 export type CigarOp = {
-  op: string;  // M, I, D, N, S, H, P, =, X
+  op: CigarSymbol;
   length: number
 }
+
 
 export type Strand = '-' | '+';
 
--- a/src/main/PileupCache.js
+++ b/src/main/PileupCache.js
@@ -28,7 +28,7 @@ export type VisualAlignment = {
 };
 
 // This is typically a read pair, but may be a single read in some situations.
-type VisualGroup = {
+export type VisualGroup = {
   key: string;
   row: number;  // pileup row.
   span: ContigInterval<string>;  // tip-to-tip span for the read group
--- a/src/main/PileupCache.js
+++ b/src/main/PileupCache.js
@@ -24,6 +24,7 @@ export type VisualAlignment = {
   strand: Strand;
   refLength: number;  // span on the reference (accounting for indels)
   mismatches: Array<BasePair>;
+  ops: Object[];
 };
 
 // This is typically a read pair, but may be a single read in some situations.
--- a/src/main/PileupTrack.js
+++ b/src/main/PileupTrack.js
@@ -7,7 +7,7 @@
 import type {Strand, Alignment, AlignmentDataSource} from './Alignment';
 import type {TwoBitSource} from './TwoBitDataSource';
 import type {BasePair} from './pileuputils';
-import type {VisualAlignment} from './PileupCache';
+import type {VisualAlignment, VisualGroup} from './PileupCache';
 
 var React = require('./react-shim'),
     d3 = require('d3'),
@@ -20,7 +20,8 @@ var React = require('./react-shim'),
     ContigInterval = require('./ContigInterval'),
     Interval = require('./Interval'),
     DisplayMode = require('./DisplayMode'),
-    PileupCache = require('./PileupCache');
+    PileupCache = require('./PileupCache'),
+    dataCanvas = require('./data-canvas');
 
 
 var READ_HEIGHT = 13;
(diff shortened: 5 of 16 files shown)


armish:
Reviewed 2 of 4 files at r1, 14 of 14 files at r2.

---

<sup>**[scripts/lint.sh, line 2 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/257#-Jzr0Yeu3CVo_9KG2Jv0)** ([raw file](https://github.com/hammerlab/pileup.js/blob/81addeb6f0edc37b1e759f41075bdfa1ecde671d/scripts/lint.sh#L2)):</sup>
short-term or long-term disabling  for the checks?

---

<sup>**[src/main/data-canvas.js, line 142 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/257#-Jzr1sS1k91y3Mw7h8sG)** ([raw file](https://github.com/hammerlab/pileup.js/blob/81addeb6f0edc37b1e759f41075bdfa1ecde671d/src/main/data-canvas.js#L142)):</sup>
never thought about this before, but are text styles/fonts going to cause us trouble? (e.g. tricky fonts that context doesn't know how to draw, etc.)

---

<sup>**[src/test/data-canvas-test.js, line 60 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/257#-Jzr-iZiU2j9B1RK63Ho)** ([raw file](https://github.com/hammerlab/pileup.js/blob/81addeb6f0edc37b1e759f41075bdfa1ecde671d/src/test/data-canvas-test.js#L60)):</sup>
hmm, a little bit puzzled by this: do you have to know a click's coordinates before drawing an object to be able to capture the hits?

---

<sup>**[src/test/PileupTrack-test.js, line 9 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/257#-Jzr0JKLiYeLIMFP1Wad)** ([raw file](https://github.com/hammerlab/pileup.js/blob/81addeb6f0edc37b1e759f41075bdfa1ecde671d/src/test/PileupTrack-test.js#L9)):</sup>
leftover?

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/257)

danvk:
Addressed all comments & added fonts/changing to rects for mismatches. PTAL.

---

Review status: 14 of 16 files reviewed at latest revision, 4 unresolved discussions, some commit checks pending.

---

<sup>**[scripts/lint.sh, line 0 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/257#-Jzr0Yeu3CVo_9KG2Jv0)** ([raw file](https://github.com/hammerlab/pileup.js/blob/81addeb6f0edc37b1e759f41075bdfa1ecde671d/scripts/lint.sh#L0)):</sup>
Long-term. Unless we can find a good way to get this sort of thing to type-check with Flow. But the classes are all so dynamic. It's hard to see how Flow could possibly keep track of what we're doing here.

---

<sup>**[src/main/data-canvas.js, line 0 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/257#-Jzr1sS1k91y3Mw7h8sG)** ([raw file](https://github.com/hammerlab/pileup.js/blob/81addeb6f0edc37b1e759f41075bdfa1ecde671d/src/main/data-canvas.js#L0)):</sup>
`CanvasRenderingContext2D` has a [`font`][1] property which can be set to any CSS font string. I've update this code to show a letter or rectangle as is appropriate for the scale. Check it out.

It would be nice if we could figure out what this font string should be using CSS, but I'd like to punt on that for a future PR.

[1]: https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/font

---

<sup>**[src/test/data-canvas-test.js, line 0 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/257#-Jzr-iZiU2j9B1RK63Ho)** ([raw file](https://github.com/hammerlab/pileup.js/blob/81addeb6f0edc37b1e759f41075bdfa1ecde671d/src/test/data-canvas-test.js#L0)):</sup>
Yep. It _feels_ inefficient, until you remember that clicks don't happen that often and that rendering a scene is really cheap if you don't actually have to render any pixels.

---

<sup>**[src/test/PileupTrack-test.js, line 9 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/257#-Jzr0JKLiYeLIMFP1Wad)** ([raw file](https://github.com/hammerlab/pileup.js/blob/81addeb6f0edc37b1e759f41075bdfa1ecde671d/src/test/PileupTrack-test.js#L9)):</sup>
Yep, nice catch! Removed.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/257)

danvk:
@ihodes would enjoy a review from you, too, since this is introducing a new abstraction.

armish:
Reviewed 2 of 2 files at r3.

---

<sup>**[src/test/data-canvas-test.js, line 60 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/257#-Jzr-iZiU2j9B1RK63Ho)** ([raw file](https://github.com/hammerlab/pileup.js/blob/81addeb6f0edc37b1e759f41075bdfa1ecde671d/src/test/data-canvas-test.js#L60)):</sup>
I see -- that does make sense then. Thanks a lot for explaining the rationale!

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/257)

ihodes:
Yeah, this approach is pretty great. Nicely done.

---

---

<sup>**[src/main/data-canvas.js, line 17 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/257#-JzulxBf5QMSyDSce5ZH)** ([raw file](https://github.com/hammerlab/pileup.js/blob/42d242bd8b5d82cd4337303b25283e896086bc37/src/main/data-canvas.js#L17)):</sup>
FWIW I believe you can give an optional parameter a default value in the type signature, like `... onlyAccessors: ?boolean = false) {...`

---

<sup>**[src/main/data-canvas.js, line 25 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/257#-JzvCyOdBuGiPqzJyohj)** ([raw file](https://github.com/hammerlab/pileup.js/blob/42d242bd8b5d82cd4337303b25283e896086bc37/src/main/data-canvas.js#L25)):</sup>
Just a note. No idea if this'll actually be a bottleneck, but in my deep-dive into JS JITs last year, I found that getters and setters in JS (at least then) could only be handled in the interpreter/could never get JITted. So this could potentially cause some serious slowness, depending. 

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/257)

danvk:

---

<sup>**[src/main/data-canvas.js, line 17 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/257#-JzulxBf5QMSyDSce5ZH)** ([raw file](https://github.com/hammerlab/pileup.js/blob/42d242bd8b5d82cd4337303b25283e896086bc37/src/main/data-canvas.js#L17)):</sup>
Nice idea, but it looks like our transpiler doesn't support that yet:

```
function forward(obj        , target        , onlyAccessors           = false) {
Uncaught SyntaxError: Unexpected token =
```

---

<sup>**[src/main/data-canvas.js, line 25 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/257#-JzvCyOdBuGiPqzJyohj)** ([raw file](https://github.com/hammerlab/pileup.js/blob/42d242bd8b5d82cd4337303b25283e896086bc37/src/main/data-canvas.js#L25)):</sup>
I haven't noticed any slowdown from using data-canvas, but I'll keep that in mind. I think the cost of throwing lots of pixels on the canvas dwarfs the JavaScript costs.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/257)

danvk:
Thanks! I'll go ahead and merge.

---

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/257)

PR #269, opened 2015-09-25
Remove and ignore some modules to speed up Flow

danvk:
See https://github.com/facebook/flow/issues/863.

This improved the speed of an incremental `flow status` from ~2.5s on my machine to ~0.5s.

@armish @ihodes After this is merged, I'd strongly recommend deleting your `node_modules` directory and re-running `npm install` to get rid of all the deps which are no longer used.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/269)


armish:
looks great! I was surprised to see the new version of flow running that fast, but this is even better :dash:

Issue #273, opened 2015-09-25
Improve read pop-up


ihodes:
Would be nice to have some basic, nice presentation of the metadata on each read object.

![small popup box on firefox](https://cloud.githubusercontent.com/assets/167803/10108702/789c91ec-6390-11e5-9e1e-0ded3e6ab79d.png)

![ugly box on chrome](https://cloud.githubusercontent.com/assets/167803/10108710/857a467a-6390-11e5-8312-1e3847ba6a95.png)

danvk:
What, you're not satisfied with a JSON dump? :)

PR #277, opened 2015-09-25
Migrate VCFTrack from SVG→canvas

### LINKED ISSUE #255: D3/SVG → canvas

danvk:
Fixes #255 

I took the opportunity to factor out some code into a shared `canvas-utils` module.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/277)


src/main/GenomeTrack.js
```diff
@@ -27,7 +28,7 @@ var GenomeTrack = React.createClass({
     onRangeChange: React.PropTypes.func.isRequired,
   },
   render: function(): any {
-    return <div><canvas ref='canvas' /></div>;
+    return <canvas />;
   },
   componentDidMount: function() {
     var div = this.getDOMNode();
```
ihodes: Did not know you could do this, nice. (As an aside, are you using `div` here? (It's also `canvas` now (sorry for chiming in))).

danvk: You can in old-style React components (i.e. `React.createClass`) but not in new-style ES6 classes. With those you have to do `React.findDOMNode(this)`.

danvk: (also, it was dead code -- will remove)

danvk:
This revealed another advantage of using canvas: previously it looked like the VariantTrack was being tested, but it really wasn't (the d3 methods were being called with empty arrays). With canvas, the drawing code clearly wasn't being run. So I added a test.

armish:
just two minor comments -- but looks great! :+1:

Kudos for refactoring.

---

Reviewed 7 of 8 files at r1, 3 of 3 files at r2.

---

<sup>**[src/main/canvas-utils.js, line 2 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/277#-K-5AbXns1ZvhtIm0l_0)** ([raw file](https://github.com/hammerlab/pileup.js/blob/1c183f4f2e38cb85a7e97d2a34dfc55c6571e7f2/src/main/canvas-utils.js#L2)):</sup>
Thanks for factoring these out!

---

<sup>**[src/main/VariantTrack.js, line 56 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/277#-K-5BR-VzrkLNEgEQm5d)** ([raw file](https://github.com/hammerlab/pileup.js/blob/1c183f4f2e38cb85a7e97d2a34dfc55c6571e7f2/src/main/VariantTrack.js#L56)):</sup>
is it worth refactoring this into: `{width, height} = this.props` to make it consistent across tracks?

---

<sup>**[src/main/VariantTrack.js, line 103 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/277#-K-5BjLexXSXFIt4XdKe)** ([raw file](https://github.com/hammerlab/pileup.js/blob/1c183f4f2e38cb85a7e97d2a34dfc55c6571e7f2/src/main/VariantTrack.js#L103)):</sup>
I know this is unrelated, but maybe it is better if we `console.log` or `console.dir` this one. IMHO alert behavior is pretty annoying especially with lines and lines of JSON in it ;)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/277)

danvk:
Thanks for the review -- will merge on green.

---

Review status: 9 of 10 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

<sup>**[src/main/VariantTrack.js, line 56 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/277#-K-5BR-VzrkLNEgEQm5d)** ([raw file](https://github.com/hammerlab/pileup.js/blob/1c183f4f2e38cb85a7e97d2a34dfc55c6571e7f2/src/main/VariantTrack.js#L56)):</sup>
Done.

---

<sup>**[src/main/VariantTrack.js, line 103 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/277#-K-5BjLexXSXFIt4XdKe)** ([raw file](https://github.com/hammerlab/pileup.js/blob/1c183f4f2e38cb85a7e97d2a34dfc55c6571e7f2/src/main/VariantTrack.js#L103)):</sup>
Yeah, this needs to be rethought. See #273.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/277)

armish:
Reviewed 1 of 1 files at r3.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/277)

Issue #278, opened 2015-09-28
Cache getDataContext calls


danvk:
I see them taking up ~5% of CPU cycles in a profile when I pan around wildly. This makes them the single biggest hog:

![screen shot 2015-09-28 at 1 30 04 pm](https://cloud.githubusercontent.com/assets/98301/10143248/25ca7842-65e5-11e5-9386-6204501d7a42.png)

Caching the DataContext in each visualization should eliminate all those calls to `forward`. Alternatively, I could modify `dataCanvas.getDataContext` to do the caching itself.

PR #279, opened 2015-09-28
Cache calls to getDataContext

danvk:
See #278 

This seems to be a small performance win.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/279)

### DIFF
--- a/src/main/data-canvas.js
+++ b/src/main/data-canvas.js
@@ -40,14 +40,25 @@ function DataContext(ctx: CanvasRenderingContext2D) {
 
 var stubGetDataContext = null;
 
-// This is exposed as a method for easier testing.
+/**
+ * Get a DataContext for the built-in CanvasRenderingContext2D.
+ *
+ * This caches DataContexts and facilitates stubbing in tests.
+ */
 function getDataContext(ctx: CanvasRenderingContext2D) {
   if (stubGetDataContext) {
     return stubGetDataContext(ctx);
   } else {
-    return new DataContext(ctx);
+    for (var i = 0; i < getDataContext.cache.length; i++) {
+      var pair = getDataContext.cache[i];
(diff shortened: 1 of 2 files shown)


armish:
optimize mercilessly ;) 

I just added a question to hear your words of wisdom on caching style, but this looks great!


---

Reviewed 2 of 2 files at r1.

---

<sup>**[src/main/data-canvas.js, line 61 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/279#-K-K5AA4bUGaNMfnhcKt)** ([raw file](https://github.com/hammerlab/pileup.js/blob/7be44363788f1f8bb27d5aefaec21c3eb6119e17/src/main/data-canvas.js#L61)):</sup>
What is the advantage of doing this over defining a global variable similar to what you have done for `stubGetDataContext` a few lines above this?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/279)

danvk:

---

<sup>**[src/main/data-canvas.js, line 61 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/279#-K-K5AA4bUGaNMfnhcKt)** ([raw file](https://github.com/hammerlab/pileup.js/blob/7be44363788f1f8bb27d5aefaec21c3eb6119e17/src/main/data-canvas.js#L61)):</sup>
It doesn't make a big difference either way. The other "globals" are actually scoped to the `data-canvas` module, so other modules can't access them. I suppose that using `getDataContext.cache` means that the `cache` object is exported, so maybe that's a negative. On the other hand, it's more obvious what it's a cache for.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/279)

PR #280, opened 2015-09-28
Run envify as part of build

### LINKED ISSUE #259: Ensure that React is not running in dev mode
Some of the call stacks in pileup.js are extremely deep:

screen shot 2015-09-22 at 4 18 44 pm

This makes me wonder if we're using a dev version of react, rather than the "production" version.

danvk:
Fixes #259 

Exporting `NODE_ENV=production` changes which version of React is used. This results in a smaller JS bundle and fewer runtime checks (=faster).

After this, running `npm run watch` will give you a debug build. Running `npm run build` will give you a "production" build.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/280)


armish:
:shipit: 


---

Reviewed 2 of 2 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/280)

PR #285, opened 2015-09-29
Support retina displays

### LINKED ISSUE #261: Use high-resolution canvas on retina displays
canvas tracks are currently low-res on retina displays

danvk:
Fixes #261 

See [this article][1]. The bit about `backingStoreRatio` seems to be out-of-date. I tested this in Firefox & Chrome.

[1]: http://www.html5rocks.com/en/tutorials/canvas/hidpi/

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/285)

### DIFF
--- a/src/main/d3utils.js
+++ b/src/main/d3utils.js
@@ -34,17 +34,22 @@ function formatRange(viewSize: number): any {
 }
 
 /**
- * Set attributes on an element. This can be used in place of, e.g.
- * d3.select(div).attr({width, height});
+ * Sizes a canvas appropriately for this device.
  */
-function setAttributes(el: Element, attrs: {[key:string]: any}) {
-  for (var k in attrs) {
-    el.setAttribute(k, attrs[k]);
+function sizeCanvas(el: HTMLCanvasElement, width: number, height: number) {
+  var ratio = window.devicePixelRatio;
+  el.width = width * ratio;
+  el.height = height * ratio;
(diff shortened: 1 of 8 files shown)


src/main/d3utils.js
```diff
@@ -34,17 +34,22 @@ function formatRange(viewSize: number): any {
...
+ * Sizes a canvas appropriately for this device.
  */
-function setAttributes(el: Element, attrs: {[key:string]: any}) {
-  for (var k in attrs) {
-    el.setAttribute(k, attrs[k]);
+function sizeCanvas(el: HTMLCanvasElement, width: number, height: number) {
+  var ratio = window.devicePixelRatio;
+  el.width = width * ratio;
+  el.height = height * ratio;
+  el.style.width = width + 'px';
```
ihodes: Why do you need to set both (style and directly on the element)?

ihodes:
LGTM, one question

src/main/d3utils.js
danvk: Notice that they're set to different things. On a retina display, for example, `ratio=2`. So you might have `canvas.width=800` whereas `canvas.style.width="400px"`. This ensures that there are the right number of pixels to match the resolution of the screen.

ihodes: Ah right, makes sense.

Issue #268, opened 2015-09-25
Flaky test on Travis


danvk:
We've been getting occasional [broken builds][1], always with errors looking like this:

```
  1) CoverageTrack should create correct labels for coverage:
     Uncaught Error: Invariant Violation: findComponentRoot(..., .0.1:$0.1.0.0.0): Unable to find element. This probably means the DOM was unexpectedly mutated (e.g., by the browser), usually due to forgetting a <tbody> when using tables, nesting tags like <form>, <p>, or <a>, or using non-SVG elements in an <svg> parent. Try inspecting the child nodes of the element with React ID ``. (http://localhost:8081/dist/tests.js:57524)
  
  2) GA4GHAlignment should match SamRead:
     Uncaught Error: Invariant Violation: findComponentRoot(..., .1.1:$1.1.0.0.0): Unable to find element. This probably means the DOM was unexpectedly mutated (e.g., by the browser), usually due to forgetting a <tbody> when using tables, nesting tags like <form>, <p>, or <a>, or using non-SVG elements in an <svg> parent. Try inspecting the child nodes of the element with React ID ``. (http://localhost:8081/dist/tests.js:57524)
  
  3) GeneTrack "before each" hook for "should fetch alignments from a server":
     the string "You forgot to call RecordingContext.reset()" was thrown, throw an Error :)
```

Rerunning the test usually fixes the problem. I've never seen this happen when I run the tests on the command line or in the browser.

[1]: https://travis-ci.org/hammerlab/pileup.js/builds/82184170

ihodes:
Are the tests running in Phantom? I've definitely seen race condition-y bugs I could never reproduce in a browser.

danvk:
Yeah, they are.

PR #289, opened 2015-09-30
Pull in data-canvas from NPM and add pileup.destroy()

danvk:
The data-canvas change surfaced #268 more aggressively in Chrome on my machine. I believe the source of the flakiness was that some event listeners (e.g. TwoBitDataSource.newdata) were being triggered _after_ the pileup was removed from the DOM. This led to code running in our React components after they were detached, which confused React and led to the `Invariant Violation` errors. I added a `destroy` method which silences the event listeners and detaches the React components from the DOM. This de-flaked the tests on my machine, and hopefully on Travis!

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/289)


armish:
all look good -- just a minor comment to double-check a flow behavior.


---

Reviewed 18 of 18 files at r1.

---

<sup>**[.flowconfig, line 25 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/289#-K-UobgD1RQ12j2wG4gX)** ([raw file](https://github.com/hammerlab/pileup.js/blob/e463329e508b1cec18c29908e33d29db4fc5d5c9/.flowconfig#L25)):</sup>
the old version of flow was not smart enough to override the ignore property and  include a single type file within an ignored directory. Just wanted to make sure that is not the case here as otherwise we will be missing all data-canvas type checks ;)


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/289)

danvk:

---

<sup>**[.flowconfig, line 25 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/289#-K-UobgD1RQ12j2wG4gX)** ([raw file](https://github.com/hammerlab/pileup.js/blob/e463329e508b1cec18c29908e33d29db4fc5d5c9/.flowconfig#L25)):</sup>
I changed a `lineTo` to `lineMoo` in a method that took a `DataCanvasRenderingContext2D` and got an error, so I think we're good. It is a little sad that we've both developed an instinct to second-guess Flow :/


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/289)

PR #299, opened 2015-10-08
Turn lint checks back on and fix code accordingly

armish:
Looks like during our recent refactoring (#205), lint paths become obsolete (https://github.com/hammerlab/pileup.js/blame/master/scripts/lint.sh#L9) and unfortunately, lint didn't complain about these missing folders. So essentially we were skipping lint checks altogether. This turns them back on and fixes the code problems to comply with lint checks.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/299)


src/main/PileupTrack.js
```diff
@@ -293,6 +293,7 @@ class PileupTrack extends React.Component {
     var trackingCtx = new dataCanvas.ClickTrackingContext(ctx, x, y);
     this.renderScene(trackingCtx);
     var vRead = trackingCtx.hit && trackingCtx.hit[0];
+    var alert = window.alert | console.log;
```
danvk: I think you mean `||` here (surprised flow doesn't complain about this, actually)

src/main/pileup.js
```diff
@@ -79,13 +79,13 @@ function create(elOrId: string|Element, params: PileupParams): Pileup {
                          initialRange={params.range} />, el);
   return {
     setRange(range: GenomeRange) {
-      if (reactElement == null) {
+      if (reactElement === null) {
```
danvk: It's OK here (since `reactElement` is explicitly set to `null` below), but I do want to call out that this change isn't a no-op. `x == null` will be true if `x` is either `null` or `undefined`, but `x === null` will only be true if it's null.

src/test/PileupCache-test.js
```diff
@@ -56,11 +55,6 @@ function makeReadPair(range1: ContigInterval<string>, range2: ContigInterval<str
   ];
 }
 
-function makeUnpairedRead(range: ContigInterval<string>, strand: Strand): Alignment {
```
danvk: :tada:

src/main/PileupTrack.js
danvk: The test failed due to Flow errors, so maybe it did catch it :)

danvk:
Thanks for tracking this down @armish! Just a few comments and you're good to go.

src/main/pileup.js
armish: gotcha! I changed it only because jshint was complaining about it, but good to know the dynamics behind this.

armish:
Thanks for catching those glitches, @danvk. Updated the PR according to your comments and all tests pass now.

PR #302, opened 2015-10-15
Use a mini-d3

### LINKED ISSUE #275: Sever d3 dependency
d3 accounts for ~150k of the ~500k pileup bundle:

```
$ ls -lh node_modules/d3/d3*
-rw-r--r--  1 danvk  staff   329K Jul  3 23:00 node_modules/d3/d3.js
-rw-r--r--  1 danvk  staff   148K Jul  3 23:00 node_modules/d3/d3.min.js
```

Now that we're migrating to canvas (see #255), we barely use D3 at all.

Remaining uses:

- [x] d3.select(...).attr(...)
- [ ] d3.scale (widely used)
- [ ] d3.format (for scale & location tracks)
- [ ] d3.behavior.drag

Only the last of these will be tricky to replace. It would be nice if we could depend on only a small part of d3!

danvk:
Fixes #275 

    587942 pileup.min.js before
    475250 pileup.min.js after

→ 112k savings (19%)

Seems like a nice win. What do you guys think of the approach?

@ihodes @armish

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/302)

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,6 +25,7 @@
     "url": "https://github.com/hammerlab/pileup.js.git"
   },
   "scripts": {
+    "minid3": "./scripts/make-mini-d3.sh",
     "jstransform": "./scripts/jstransform.sh",
     "jstransform-watch": "./scripts/jstransform.sh --watch",
     "browserify-all": "npm run browserify && npm run browserify-test",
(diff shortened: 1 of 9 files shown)


armish:
Looks like Mike Bostock is at it again with that `smash` utility! Seems like a nice hack to reduce the library size even further. I would say go for it until we can completely get rid of d3.

ihodes:
I think it's reasonable enough, nice savings for sure.

danvk:
Yeah, `SMASH` might have made sense in 2011 but it has no purpose now. They plan to use ES6 `import` statements in D3 4.0 https://github.com/mbostock/d3/issues/2220

PR #303, opened 2015-10-15
Allow any track to be the reference

danvk:
The issue was that I was calling `_.findWhere`, which expects key-value pairs to look for, rather than `_.find`, which takes a predicate.

Why Flow didn't catch this is beyond me.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/303)


armish:
:+1: looks great!


---

Reviewed 4 of 4 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/303)

PR #309, opened 2015-10-16
Generate a source map for pileup.min.js

danvk:
To facilitate debugging bundle size with source-map-explorer

The net result here is that we get two new files in `dist`:

```
dist/pileup.js.map
dist/pileup.min.js.map
```

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/309)


armish:
LGTM. It is sad that `uglify` cannot directly use the map `browserify` generates :(

danvk:
Yeah. I'm going to play around with [uglifyify][1] and see if we can get some better results.

[1]: https://github.com/hughsk/uglifyify

PR #310, opened 2015-10-16
Be more aggressive about minification

### LINKED ISSUE #307: Use minified React
React [currently takes up][1] ~207k in our minified JS bundle.

But the pre-built files that Facebook distributes are significantly smaller than this:

    129K react-with-addons.min.js
    119K  react.min.js

We should use these instead. This may require removing the react shim (#246).

[1]: https://cdn.rawgit.com/danvk/source-map-explorer/1f02ba07a2d05c7c7dc0027d31c257b12ffe3c8f/demo.html

danvk:
There are two things going on here:

* Using `-g [ envify ]` instead of `-t [envify]` applies the transform globally (i.e. including to code in `node_modules`), rather than just to our code. This means that it sets the `NODE_ENV` variable inside of React.
* The [uglifyify][1] transform minifies each module as browserify sees it. This means that dead `require` statements are removed.

These combine to shrink React pretty dramatically, since all the dev-mode code can be stripped.

```
452,690 pileup.min.js before
394,079 pileup.min.js after
```

∆ = 58,611 bytes (13%)

Looking at the source map, most of the difference is coming from React:

```
207,354 /react before (172 files in bundle)
139,935 /react after (167* files in bundle)
```

∆ = 67,419 bytes (33%)

Closes #307 

\* the dropped files are:

    node_modules/react/lib/ReactDefaultPerf.js
    node_modules/react/lib/ReactDefaultPerfAnalysis.js
    node_modules/react/lib/ReactTestUtils.js
    node_modules/react/lib/performance.js
    node_modules/react/lib/performanceNow.js

[1]: https://github.com/hughsk/uglifyify

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/310)


danvk:
cc @ihodes

package.json
```diff
@@ -29,10 +29,10 @@
     "jstransform": "./scripts/jstransform.sh",
     "jstransform-watch": "./scripts/jstransform.sh --watch",
     "browserify-all": "npm run browserify && npm run browserify-test",
-    "browserify": "browserify -r .:pileup -v -t [ envify --NODE_ENV production ] --debug -o dist/pileup.js",
+    "browserify": "browserify -r .:pileup -v -g [ envify --NODE_ENV production ] -g uglifyify --debug -o dist/pileup.js",
```
armish: does this lead to double uglification (first via uglifyify and then directly via uglify)?

armish:
File size looks remarkably small now! Thanks for testing these options out, @danvk.

package.json
danvk: Yes -- though this is actually recommended!

> When using uglifyify you should generally also use Uglify, to achieve the smallest output. Uglifyify provides an additional optimization when used with Uglify, but does not provide all of the optimization that using Uglify on its own does, so it's not a replacement.

armish: aha, magic :crystal_ball:

PR #311, opened 2015-10-16
Show center lines on pileups

### LINKED ISSUE #180: The column containing your selected variant should be highlighted

danvk:
pileup-center-line

Fixes #180

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/311)

### DIFF
--- a/src/main/LocationTrack.js
+++ b/src/main/LocationTrack.js
@@ -55,10 +55,10 @@ class LocationTrack extends React.Component {
         leftLineX = scale(midPoint);
 
     // Left line
-    canvasUtils.drawLine(ctx, rightLineX, 0, rightLineX, height);
+    canvasUtils.drawLine(ctx, leftLineX, 0, leftLineX, height);
 
     // Right line
-    canvasUtils.drawLine(ctx, leftLineX, 0, leftLineX, height);
+    canvasUtils.drawLine(ctx, rightLineX, 0, rightLineX, height);
 
     // Mid label
     var midLabelFormat = d3.format(',d'),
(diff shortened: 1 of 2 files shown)


armish:
This looks really nice! :ship: 

As a side question: Do you think we should make these kinds of enrichments optional? (see also #312)


---

Reviewed 2 of 2 files at r1.

---

<sup>**[src/main/LocationTrack.js, line 58 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/311#-K0mg2-62bre6bsxl6LB)** ([raw file](https://github.com/hammerlab/pileup.js/blob/16ae718578888d086fb1cd8868fafb327b64d547/src/main/LocationTrack.js#L58)):</sup>
good catch! :)


---

<sup>**[src/main/PileupTrack.js, line 307 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/311#-K0mgMnFF9Yav0Mv61x7)** ([raw file](https://github.com/hammerlab/pileup.js/blob/7c5f52677f694e8f71b6a3bf163ba6d64710ed0f/src/main/PileupTrack.js#L307)):</sup>
nice feature! Maybe I should also add this to the location track.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/311)

danvk:
Thanks!

My bias is usually against adding options until there's a demonstrated need for them, since it's easy to wind up with [huge sets][1] of [options][2] which can interact with each other in unexpected ways.

But yeah, for some high level things (e.g. showing reads as pairs) I think options are a good idea.

[1]: http://dygraphs.com/options.html
[2]: https://developers.google.com/chart/interactive/docs/gallery/linechart?hl=en#configuration-options

Issue #306, opened 2015-10-15
Coverage track skips when panning


ihodes:
![screen recording 2015-10-15 at 06 42 pm](https://cloud.githubusercontent.com/assets/167803/10529496/5ae8e362-736c-11e5-900b-627b9e46cb6e.gif)

danvk:
Looks like it needs to round down to include the previous bin.

PR #312, opened 2015-10-16
Show mismatches in the coverage track

### LINKED ISSUE #233: Show mismatches in CoverageTrack
The coverage track can give a quick overview of how strong the evidence for a variant is.

armish:
fixes #233.

screen shot 2015-10-16 at 5 57 33 pm

screen shot 2015-10-16 at 5 56 54 pm

screen shot 2015-10-16 at 5 54 58 pm

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/312)

### DIFF
--- a/src/main/types.js
+++ b/src/main/types.js
@@ -43,3 +43,15 @@ export type InflatedBlock = {
   compressedLength: number;
   buffer: ArrayBuffer;
 }
+
+// Coverage Track
+export type BinSummary = {
+  count: number;
+  mismatches: string[];
+}
+
+export type BinSummaryWithLocation = {
+  position: string;
+  count:number;
+  mismatches: string[];
+}
--- a/lib/underscore.js
+++ b/lib/underscore.js
@@ -31,7 +31,7 @@ declare module "underscore" {
 
   declare function sortBy<T>(a: T[], iteratee: (val: T)=>any): T[];
 
-  declare function filter<T>(o: {[key:string]: T}, pred: (val: T, k: string)=>boolean): T[];
+  declare function filter<T>(o: {[key:string]: T} | Array<T>, pred: (val: T, k: string)=>boolean): T[];
 
   declare function isEmpty(o: any): boolean;
 
@@ -43,4 +43,3 @@ declare module "underscore" {
   declare function values<T>(o: {[key: any]: T}): T[];
   declare function flatten(a: Array<any>): Array<any>;
 }
-
--- a/src/main/CoverageTrack.js
+++ b/src/main/CoverageTrack.js
@@ -6,6 +6,7 @@
 
 import type * as SamRead from './SamRead';
 import type * as Interval from './Interval';
+import type {TwoBitSource} from './TwoBitDataSource';
 
 var React = require('./react-shim'),
     d3 = require('d3/minid3'),
@@ -14,6 +15,7 @@ var React = require('./react-shim'),
     d3utils = require('./d3utils'),
     _ = require("underscore"),
     dataCanvas = require('data-canvas'),
+    {getOpInfo} = require('./pileuputils'),
     style = require('./style'),
     ContigInterval = require('./ContigInterval');
 
@@ -21,23 +23,42 @@ var React = require('./react-shim'),
 /**
  * Extract summary statistics from the read data.
  */
-function extractSummaryStatistics(reads: Array<SamRead>, contig: string) {
-  var binCounts = ({}: {[key: number]: number});
+function extractSummaryStatistics(reads: Array<SamRead>,
+                                  contig: string,
+                                  referenceSource: TwoBitSource) {
+  var binCounts = ({}: {[key: number]: any});
 
   // This is written in an imperative style (instead of with _.groupBy)
   // as an optimization.
   _.each(reads, read => {
     var interval = read.getInterval();
 
+    var opInfo = getOpInfo(read, referenceSource);
+
     var start = interval.start(),
         stop = interval.stop();
     for (var j = start; j <= stop; j++) {
-      binCounts[j] = (binCounts[j] || 0) + 1;
+      var binCount = binCounts[j] || { count: 0, mismatches: [] };
+      binCount.count += 1;
+      binCounts[j] = binCount;
     }
-  });
-  var maxCoverage = _.max(binCounts);
 
-  var posCounts = _.map(binCounts, (count, position) => ({position: Number(position), count}));
+    // Capture mismatches for future use
+    _.each(opInfo.mismatches, m => {
+      var binCount = binCounts[m.pos + 1];
+      if(binCount) {
+        binCounts[m.pos + 1].mismatches.push(m.basePair);
+      } else {
+        // do nothing, we don't info on this position yet
+      }
+    });
+  });
+  var maxCoverage = _.max(binCounts, bc => bc.count).count;
+
+  var posCounts = _.map(binCounts,
+                        (bc, position) => ({position: Number(position),
+                                            count: bc.count,
+                                            mismatches: bc.mismatches}));
   var sortedPosCounts = _.sortBy(posCounts, bin => bin.position);
 
   return {binCounts: sortedPosCounts, maxCoverage};
@@ -65,17 +86,20 @@ class CoverageTrack extends React.Component {
   }
 
   componentDidMount() {
-    this.props.source.on('newdata', () => {
+    var updateState = () => {
       var ci = new ContigInterval(this.props.range.contig, 0, Number.MAX_VALUE),
           reads = this.props.source.getAlignmentsInRange(ci),
-          {binCounts, maxCoverage} = extractSummaryStatistics(reads, this.props.range.contig);
+          {binCounts, maxCoverage} = extractSummaryStatistics(reads, this.props.range.contig, this.props.referenceSource);
 
       this.setState({
         reads,
         binCounts,
         maxCoverage
       });
-    });
+    };
+
+    this.props.source.on('newdata', updateState);
+    this.props.referenceSource.on('newdata', updateState);
   }
 
   componentDidUpdate(prevProps: any, prevState: any) {
@@ -85,7 +109,7 @@ class CoverageTrack extends React.Component {
     }
   }
 
-  binsInRange(): Array<{position:string,count:number}> {
+  binsInRange(): Array<{position:string,count:number,mismatches:Array<string>}> {
     var {start, stop} = this.props.range;
     return this.state.binCounts.filter(
         ({position}) => (position >= start && position <= stop));
@@ -122,7 +146,6 @@ class CoverageTrack extends React.Component {
     ctx.reset();
     ctx.clearRect(0, 0, ctx.canvas.width, ctx.canvas.height);
 
-    ctx.fillStyle = style.COVERAGE_BIN_COLOR;
     var barWidth = xScale(1) - xScale(0),
         barPadding = barWidth * style.COVERAGE_BIN_PADDING_CONSTANT;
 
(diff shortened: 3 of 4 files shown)


danvk:
Woohoo! I think IGV only shows a mismatch if it's above some threshold. Can
you match that?

armish:
done (current threshold is 1 read).

Will make this configurable by the user through the track object (an issue to follow)

danvk:
It looks like it's actually a [percentage threshold][1]:

> If a nucleotide differs from the reference sequence in greater than 20% of quality weighted reads, IGV colors the bar in proportion to the read count of each base (A, C, G, T).

Let's just match their behavior until we have a reason to diverge.

Overall this looks great, Arman!

[1]: http://www.broadinstitute.org/igv/AlignmentData


---

Review status: 0 of 3 files reviewed at latest revision, 12 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/CoverageTrack.js, line 35 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/312#-K0qVESXEqE2xJBmPKgU)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0dde7a745ece9716d374bd930032d06c9be1c267/src/main/CoverageTrack.js#L35)):</sup>
How about pulling this out into a `type`?

```javascript
type BinSummary = {
  count: number;
  mismatches: string[];
}
```


---

<sup>**[src/main/CoverageTrack.js, line 49 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/312#-K0qWE0dsDjtd1WXIj0a)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0dde7a745ece9716d374bd930032d06c9be1c267/src/main/CoverageTrack.js#L49)):</sup>
reassigning the object on every iteration feels a little odd. How about...

```javascript
if (!binCounts[j]) binCounts[j] = {count: 0, mismatches: []};
binCounts[j].count += 1;
```


---

<sup>**[src/main/CoverageTrack.js, line 53 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/312#-K0qVkMm9rsqWhNqdwfE)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0dde7a745ece9716d374bd930032d06c9be1c267/src/main/CoverageTrack.js#L53)):</sup>
No need to change it, but FWIW arrays have their own `.forEach` method. No underscore needed.


---

<sup>**[src/main/CoverageTrack.js, line 55 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/312#-K0qWlvcrTXHc0DgJRyA)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0dde7a745ece9716d374bd930032d06c9be1c267/src/main/CoverageTrack.js#L55)):</sup>
style nit: space after `if`


---

<sup>**[src/main/CoverageTrack.js, line 56 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/312#-K0qVXEA99FBF__eDOjd)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0dde7a745ece9716d374bd930032d06c9be1c267/src/main/CoverageTrack.js#L56)):</sup>
This would be simpler/more efficient if `mismatches` were an object mapping base pair → count. Then you could drop the `countBy` in the rendering code.


---

<sup>**[src/main/CoverageTrack.js, line 118 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/312#-K0qXCLuwwsMGJRPAx63)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0dde7a745ece9716d374bd930032d06c9be1c267/src/main/CoverageTrack.js#L118)):</sup>
nit: space after `,`


---

<sup>**[src/main/CoverageTrack.js, line 121 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/312#-K0qXJQOMK3M8Eebg1FU)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0dde7a745ece9716d374bd930032d06c9be1c267/src/main/CoverageTrack.js#L121)):</sup>
May as well fix #306 while you're in here — change `start`→`start - 1` and `stop`→`stop + 1`.


---

<sup>**[src/main/CoverageTrack.js, line 172 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/312#-K0qXinDOdH0X5MO1iOn)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0dde7a745ece9716d374bd930032d06c9be1c267/src/main/CoverageTrack.js#L172)):</sup>
style nit: space after `if`


---

<sup>**[src/main/CoverageTrack.js, line 173 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/312#-K0qXh6bpRlIbqMe2rwn)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0dde7a745ece9716d374bd930032d06c9be1c267/src/main/CoverageTrack.js#L173)):</sup>
use a two-space indent


---

<sup>**[src/main/CoverageTrack.js, line 174 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/312#-K0qY3spBSr-4W186FYh)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0dde7a745ece9716d374bd930032d06c9be1c267/src/main/CoverageTrack.js#L174)):</sup>
Order these so that the most-common mismatch is on the bottom.


---

<sup>**[src/main/CoverageTrack.js, line 181 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/312#-K0qXnd2Dt4BC6gYSdL6)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0dde7a745ece9716d374bd930032d06c9be1c267/src/main/CoverageTrack.js#L181)):</sup>
also for click-tracking in the future


---

<sup>**[src/test/CoverageTrack-test.js, line 95 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/312#-K0qYFlPT9D4W_dqp2l4)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0dde7a745ece9716d374bd930032d06c9be1c267/src/test/CoverageTrack-test.js#L95)):</sup>
You can write this more succinctly as:

```javascript
expect(findMismatchBins()).to.deep.equal(
    [{position: 7500765, count: 22, base: 'T'}]);
```


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/312)

JPFinnigan:
"Let's just match their behavior until we have a reason to diverge."

One reason--w/h I have encountered multiple times in IGV--would be the situation in which passing somatic variants present at less than AF 0.20 are not highlighted in the coverage track. Perhaps this is not the most workable solution, but it would be neat if in situations in which a .vcf was supplied w/ an accompanying .bam the threshold for adding coloration to the coverage track were intelligently determined based on the minimum minor allele AF for variants where "Is Filtered out: NO" is true (e.g the minimum AF for "passing" variants. 

In all other circumstances one could default to the fixed 20% threshold...  

just my $0.02

Have a nice weekend!

danvk:
Thanks for chiming in @JPFinnigan! The point about somatic mutations with low AF is well-taken. I assume the rationale for having a threshold at all is reducing visual clutter. Maybe we should start with a threshold of >=2 variants to show a segment in the bar and see if it's too cluttered?

armish:
all done - our current threshold was > 1 mismatches, so I kept that as it is.


---

Review status: 0 of 4 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

<sup>**[src/main/CoverageTrack.js, line 53 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/312#-K0qVkMm9rsqWhNqdwfE)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0dde7a745ece9716d374bd930032d06c9be1c267/src/main/CoverageTrack.js#L53)):</sup>
not sure why but I like `_.each` better than `.forEach` whenever I have the chance.  I agree that `_` is not needed only for this method elsewhere ;) 


---

<sup>**[src/main/CoverageTrack.js, line 56 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/312#-K0qVXEA99FBF__eDOjd)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0dde7a745ece9716d374bd930032d06c9be1c267/src/main/CoverageTrack.js#L56)):</sup>
I originally thought that this way could save us from `countBy` and mapping logic when/if users don't want to show the mismatches (that is why `countBy` is within that `if` block that checks for the options.

Do you think appending pushing chars like this to an array causes us efficiency problems?  If not, I think I would like leave that logic to `_`'s `countBy` but if you have strong feelings, I am more than happy to change it.


---

<sup>**[src/main/CoverageTrack.js, line 118 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/312#-K0qXCLuwwsMGJRPAx63)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0dde7a745ece9716d374bd930032d06c9be1c267/src/main/CoverageTrack.js#L118)):</sup>
pulled this out as a type definition now.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/312)

danvk:
Two comments, then you're good to go.


---

Reviewed 1 of 3 files at r1, 2 of 2 files at r2.
Review status: 1 of 4 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/CoverageTrack.js, line 56 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/312#-K0qVXEA99FBF__eDOjd)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0dde7a745ece9716d374bd930032d06c9be1c267/src/main/CoverageTrack.js#L56)):</sup>
It is O(n) memory vs. O(1), but my feelings aren't especially strong in this case. The thing I do feel strongly about is calling `countBy` in `extractSummaryStatistics`, rather than in `visualizeCoverage`.


---

<sup>**[src/main/CoverageTrack.js, line 10 \[r3\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/312#-K10tExnHAdWHOIW1fRF)** ([raw file](https://github.com/hammerlab/pileup.js/blob/7d7fc7b9f1f4ae85e1579c9c0532c5a66571b332/src/main/CoverageTrack.js#L10)):</sup>
Since these aren't used anywhere but `CoverageTrack`, you should just define them internally to this file.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/312)

armish:
:white_check_mark: 


---

Review status: 1 of 4 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

<sup>**[src/main/CoverageTrack.js, line 56 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/312#-K0qVXEA99FBF__eDOjd)** ([raw file](https://github.com/hammerlab/pileup.js/blob/0dde7a745ece9716d374bd930032d06c9be1c267/src/main/CoverageTrack.js#L56)):</sup>
gotcha! Made this change, so `mismatches` acts as a `base`->`number` map.

Thanks for catching this and the explanation!


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/312)

danvk:
Review status: 1 of 4 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---

<sup>**[src/main/CoverageTrack.js, line 66 \[r4\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/312#-K11-CC3igBzOAo9wudc)** ([raw file](https://github.com/hammerlab/pileup.js/blob/7a4dc0ac6742869375cd6180fdca57975bb3f102/src/main/CoverageTrack.js#L66)):</sup>
I think this is clearer as:

```javascript
binCount.mismatches[m.basePair] = 1 + basecount;
```

since you don't use the auto-incremented variable again. Or even

```javascript
var mismatches = binCount.mismatches;
mismatches[m.basePair] = 1 + (mismatches[m.basePair] || 0);
```


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/312)

armish:
Review status: 1 of 4 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

<sup>**[src/main/CoverageTrack.js, line 66 \[r4\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/312#-K11-CC3igBzOAo9wudc)** ([raw file](https://github.com/hammerlab/pileup.js/blob/7a4dc0ac6742869375cd6180fdca57975bb3f102/src/main/CoverageTrack.js#L66)):</sup>
my bad -- sorry about that unclear code piece. I like the latter better, so changed this to your second suggestion.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/312)

danvk:
:shipit: LGTM


---

Review status: 1 of 4 files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/312)

src/main/types.js
```diff
@@ -43,3 +43,15 @@ export type InflatedBlock = {
   compressedLength: number;
   buffer: ArrayBuffer;
 }
+
+// Coverage Track
+export type BinSummary = {
```
danvk: Looks like these definitions snuck in after all. @armish Mind sending a quick PR to remove them?

armish: argh! Really sorry about that. See #317 for the fix.

PR #315, opened 2015-10-20
Sever dependence on d3.{scale,format}

danvk:
See #275 

This gets minid3 down to ~15k in the minified bundle, just the bit contributed by `d3.behavior.drag`.

After trying to update cycledash to use the latest pileup, I realized there was a problem. The directory structure looks like this:

```
cycledash
  node_modules
    d3
    pileup
      node_modules
      ...
```

The `make-mini-d3.sh` script expects its copy of D3 to be in `pileup/node_modules`, but if the parent project depends on d3 itself, then there's no reason to expect that this will be the case.

If we want to get the benefit of not depending on all of d3, then we need to either sever the dependency completely or distribute a copy of minid3 outside of `node_modules`.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/315)

### DIFF
--- a/package.json
+++ b/package.json
@@ -93,6 +93,7 @@
     "mocha-lcov-reporter": ">=0.0.2",
     "mocha-phantomjs": "3.5.3",
     "mocha-phantomjs-istanbul": "0.0.2",
+    "number-to-locale-string": "^1.0.0",
     "parse-data-uri": "^0.2.0",
     "phantomjs": "^1.9.17",
     "prepush-hook": "^0.1.0",
--- a/scripts/make-mini-d3.sh
+++ b/scripts/make-mini-d3.sh
@@ -8,7 +8,6 @@
 smash \
     node_modules/d3/src/start.js \
     node_modules/d3/src/format/format.js \
-    node_modules/d3/src/scale/linear.js \
     node_modules/d3/src/behavior/drag.js \
     node_modules/d3/src/end.js \
     | perl -pe 's/ε/EPSILON/g' \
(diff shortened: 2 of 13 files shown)


danvk:
```
398065 pileup.min.js before
375547 pileup.min.js after
```

→ 22k (5.6%) savings

armish:
Looks great!

PS: Unfortunate that you had to polyfill `toLocaleString` only for phantomjs :(


---

Reviewed 13 of 13 files at r1.

---

<sup>**[src/main/scale.js, line 80 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/315#-K14xBy9XMyAZYclaS4q)** ([raw file](https://github.com/hammerlab/pileup.js/blob/ec6cb71536de1306fa25dd81b9c438ba4df5c51a/src/main/scale.js#L80)):</sup>
drop this?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/315)

danvk:

---

<sup>**[src/main/scale.js, line 80 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/315#-K14xBy9XMyAZYclaS4q)** ([raw file](https://github.com/hammerlab/pileup.js/blob/ec6cb71536de1306fa25dd81b9c438ba4df5c51a/src/main/scale.js#L80)):</sup>
Done.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/315)

armish:
:ship: 


---

Reviewed 1 of 1 files at r2.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/315)

PR #316, opened 2015-10-20
Switch from jstransform --> babel

### LINKED ISSUE #295: "npm run watch" sometimes fails to update tests
I've had a hard time finding a reliable repro, but it's happened often enough that killing the task and re-running it is part of my usual debugging process.

The ideal solution here might be to switch from browserify/jstransform/watchify to webpack/babel.

danvk:
Development of jstransform has stopped. babel is the new hotness. This will let us use new ES6 features as we choose to, e.g. ES6 imports.

Using `watchify -t babelify` also lets us get rid of the separate `jstransform --watch` and `watchify` steps. Hopefully this fixes #295.

I also took the opportunity to make `dist/pileup.js` export a single `pileup` global, rather than a `require` global. This makes more sense to me. If you're including pileup in a browser, you probably want the single global `pileup` symbol. If you're going to `require` it, you probably want to require the whole pileup tree (as Cycledash does).

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/316)


armish:
looks great, @danvk! I really like that we are down to only a few simple but effective tasks/scripts for production/testing.


---

Reviewed 10 of 10 files at r1.

---

<sup>**[package.json, line 29 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/316#-K163EdYeDQeY7Z7HXcO)** ([raw file](https://github.com/hammerlab/pileup.js/blob/f16c5ef9d3158fbd14e98a6f0f8241072a2520b2/package.json#L29)):</sup>
wohoo! Clean all the tasks!


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/316)

PR #317, opened 2015-10-20
Drop left-over type definitions

armish:
These are left-overs from #312.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/317)


danvk:
LGTM, merge on green

PR #318, opened 2015-10-20
Put minid3 in src/lib

danvk:
See #275 

This will let other programs (like cycledash) depend on pileup again.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/318)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,3 +37,6 @@ dist
 /*.bed
 /*.bam
 /*.vcf
+
+# Generated files
+src/lib
--- a/package.json
+++ b/package.json
@@ -25,7 +25,6 @@
     "url": "https://github.com/hammerlab/pileup.js.git"
   },
   "scripts": {
-    "minid3": "./scripts/make-mini-d3.sh",
     "watch": "./scripts/watch.sh",
     "lint": "./scripts/lint.sh",
     "http-server": "http-server .",
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,19 +1,26 @@
 #!/bin/bash
 # Build require-ale and minified assets for distribution.
 set -o errexit
-npm run minid3  # TODO: remove
+./scripts/make-mini-d3.sh # TODO: remove
 
 # Transpile individual files. This is useful if another module,
 # e.g. cycledash, wants to require('pileup').
 # The dist/test files are required for code coverage
-babel src --out-dir dist
+babel src --ignore src/lib --out-dir dist
+cp -r src/lib dist/
 
 # Create dist/tests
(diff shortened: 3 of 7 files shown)


armish:
:+1: just two questions for clarification.


---

Reviewed 7 of 7 files at r1.

---

<sup>**[package.json, line 28 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/318#-K16DavecayspLD7fHTu)** ([raw file](https://github.com/hammerlab/pileup.js/blob/e4cc8952859c9af0b5b8726ffb527859ba9f3f07/package.json#L28)):</sup>
is there a reason why you remove this task but directly call it from other scripts?


---

<sup>**[scripts/watch.sh, line 13 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/318#-K16EhFyZgxPVX41j5EQ)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bf11eb4499bc99bdf49c09cdbc28c5cc65d62401/scripts/watch.sh#L13)):</sup>
don't you need to carry the minified d3 over to the dist, similar to the way you do in `build.sh`? I might be wrong, but if somebody calls `watch.sh` before `build.sh` that mini3d won't be there, right?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/318)

danvk:

---

<sup>**[package.json, line 28 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/318#-K16DavecayspLD7fHTu)** ([raw file](https://github.com/hammerlab/pileup.js/blob/e4cc8952859c9af0b5b8726ffb527859ba9f3f07/package.json#L28)):</sup>
Because it's not something I could ever imagine a user wanting to run directly. It's more of an intermediate step.


---

<sup>**[scripts/watch.sh, line 13 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/318#-K16EhFyZgxPVX41j5EQ)** ([raw file](https://github.com/hammerlab/pileup.js/blob/bf11eb4499bc99bdf49c09cdbc28c5cc65d62401/scripts/watch.sh#L13)):</sup>
Interestingly enough, no! These `watchify` tasks only take inputs from `src/`, not from `dist/`. In the new world of babel, the individual files in `dist/main` are only needed for distribution. The babel transforms are done inside the browserify/watchify processes.

In other words, they pull in `src/lib/minid3.js`, not `dist/lib/minid3.js`.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/318)

PR #322, opened 2015-10-21
Fix two small bugs

danvk:
These popped up when I tried updating pileup in Cycledash.

With the `scaleRange` change, we lose the invariant that the range is precisely half or double what it was before, but we gain an invariant that the center base stays fixed, which is what you notice most easily.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/322)

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -16,6 +16,9 @@
 .*node_modules/jstransformify.*
 .*node_modules/uglify.*
 .*node_modules/data-canvas.*
+.*node_modules/babel.*
+.*node_modules/d3.*
+.*node_modules/smash.*
 .*react/node_modules/fbjs.*
 .*coverage.*
 
--- a/src/main/Controls.js
+++ b/src/main/Controls.js
@@ -5,8 +5,11 @@
 'use strict';
 
 var React = require('react'),
-    types = require('./react-types'),
     _ = require('underscore');
+
+var types = require('./react-types'),
+    utils = require('./utils'),
+    Interval = require('./Interval');
 
 var Controls = React.createClass({
   propTypes: {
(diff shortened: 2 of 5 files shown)


armish:
All looks good, @danvk. Thanks for these fixes! Just one quick question/suggestion about scaleRange's behavior and it is good to go.


---

Reviewed 5 of 5 files at r1.

---

<sup>**[src/main/utils.js, line 167 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/322#-K1CQ8Nmqhqar2sA04kz)** ([raw file](https://github.com/hammerlab/pileup.js/blob/07c11c4b9459d81ff0dd9e5d28a50733128e3622/src/main/utils.js#L167)):</sup>
if `center - newSpan  / 2 < 0`, then this won't scale the range exactly by the given factor and the center will shift, right? I know it is an edge use case, but does it make sense to add those cut bases back to the stop (which will still shift the center but give you a range with expected length) or should you re-adjust the scale so that `start - center == end - start` (which will not respect the factor)?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/322)

danvk:
Review status: 3 of 5 files reviewed at latest revision, 1 unresolved discussion, some commit checks pending.

---

<sup>**[src/main/utils.js, line 167 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/322#-K1CQ8Nmqhqar2sA04kz)** ([raw file](https://github.com/hammerlab/pileup.js/blob/07c11c4b9459d81ff0dd9e5d28a50733128e3622/src/main/utils.js#L167)):</sup>
Good idea! The center was being shifted anyway (since `start` was being clamped but not `stop`), so there's really no loss in doing this. Added a test, too.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/322)

armish:
:ship: 


---

Reviewed 2 of 2 files at r2.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/322)

PR #326, opened 2015-10-22
Only render alignments which are visible

### LINKED ISSUE #325: Only render alignments which are on-screen vertically
With read depths above 20 or so, it's typical that you have to scroll vertically to see all the reads.

We only render reads which are on-screen horizontally. We should do the same thing vertically.

This would require the pileup canvas to know about its scroll state.

danvk:
Fixes #325 

This winds up being a nice optimization for deep sequences (e.g. RNAseq or WES)

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/326)

### DIFF
--- a/src/main/PileupTrack.js
+++ b/src/main/PileupTrack.js
@@ -17,6 +17,7 @@ var React = require('react'),
     d3utils = require('./d3utils'),
     {CigarOp} = require('./pileuputils'),
     ContigInterval = require('./ContigInterval'),
+    Interval = require('./Interval'),
     DisplayMode = require('./DisplayMode'),
     PileupCache = require('./PileupCache'),
     canvasUtils = require('./canvas-utils'),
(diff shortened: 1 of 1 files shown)


armish:
looks like a great optimization trick! :shipit: 

Just curious: did you observe some performance gain overall or is it just anecdotal for Cycledash data at this point? :)


---

Reviewed 1 of 1 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/326)

danvk:
It was a pretty noticeable improvement (~2x?) when I was zoomed out and panning over ranges that had been displayed previously. Both it and CoverageTrack are doing a lot of work when you're zoomed out. I have some ideas about how to improve CoverageTrack. Now that PileupTrack is faster in those cases, improvements to CoverageTrack will have a bigger effect.

I'm not totally sure I buy the FPS numbers that our frame counter reports. Sometimes it seemed like the location was updating quickly (~10fps) but the pileup was updating at a lower frequency. Will have to investigate more!

PR #327, opened 2015-10-22
Attempt to notify on code size

danvk:
We build the minified file anyway, so we may as well track it!

Notes:
- This is intentionally dependency-free. Running Python in the node Travis-CI stack is messy. I was unable to get `pip install requests` to work.
- I'll add support for changes in size once we have this running on `master`. My plan is to use the GitHub API to figure out the base for the current PR, then fetch and parse its build status.
- I had to add a `GITHUB_TOKEN` environment variable to our Travis build. I got this using Profile→Settings→Personal Access Tokens on GitHub.

References:

- [GitHub Notifications API][1]
- [Unanswered question][2] about this on StackOverflow
- [urllib2 documentation][3]
- [Travis-CI environment variables][4]

[1]: https://developer.github.com/v3/repos/statuses/
[2]: http://stackoverflow.com/questions/33244501/track-generated-file-sizes-on-github-using-travis-ci
[3]: https://docs.python.org/2/howto/urllib2.html
[4]: http://docs.travis-ci.com/user/environment-variables/#Default-Environment-Variables

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/327)

PR #328, opened 2015-10-23
Attempt to display deltas

danvk:
This builds on #327 

This is a bit roundabout because of how Travis-CI handles pull requests. As near as I can tell:

- When you push anything to GitHub, Travis-CI starts a "push" build. This has the `TRAVIS_COMMIT` environment variable set, but not `TRAVIS_PULL_REQUEST`. This makes sense, since the push might not be part of a PR.
- When you create a pull request for a branch, Travis-CI starts a "pr" build. This checks out something like `refs/pull/328/merge`, which is the PR branch as merged onto whatever its base is. This has, e.g. `TRAVIS_PULL_REQUEST=328`, and `TRAVIS_COMMIT` set to whatever the merge commit's SHA is.
- GitHub only shows statuses for the "push" commit in the PR view.

It's difficult to determine PR information in the "push" build (see [this question][1]), so instead I figure out the "push" SHA in the "pr" build. Both builds post file size information as a status on the "push" SHA, but only the "pr" build will post a file size delta. Since it typically starts after the "push" build, the better message will typically be the one that's displayed. But it is a race.

I'm not sure if there's a better way to do this. I do need to calculate sizes in "push" builds (e.g. for when you push to master, to get baselines). But unless that SO question gets a satisfactory answer, I don't see any way to reliably get the deltas to show up.

[1]: http://stackoverflow.com/questions/33307413/how-can-i-determine-which-pull-request-a-commit-belongs-to-using-the-github-api

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/328)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,7 @@ node_js:
 
 script: >
     npm run build &&
+    ./scripts/post_code_size.py dist/pileup.min.js &&
     npm run test &&
     npm run flow-check &&
     npm run lint &&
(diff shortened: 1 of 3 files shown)


ihodes:
LGTM—Cool addition!

PR #331, opened 2015-10-26
Retain lines when running babel

### LINKED ISSUE #330: Code coverage doesn't match source
See, for example, coverage for [`RemoteFile.js`][2]. The lines covered don't make logical sense. This was broken by my recent [switch to babelify][1].

The issue is that the coverage is being generated for the babel-ified source, which doesn't necessarily match up 1-1 line for line with the original source.

[1]: https://github.com/hammerlab/pileup.js/pull/316
[2]: https://coveralls.io/builds/3932994/source?filename=src%2Fmain%2FRemoteFile.js

danvk:
Fixes #330 

Compare `RemoteFile.js` coverage [before][] and [after][]. The latter makes sense, the former does not.

[before]: https://coveralls.io/builds/3932994/source?filename=src%2Fmain%2FRemoteFile.js
[after]: https://coveralls.io/builds/3953163/source?filename=src%2Fmain%2FRemoteFile.js

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/331)

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -6,7 +6,7 @@ set -o errexit
 # Transpile individual files. This is useful if another module,
 # e.g. cycledash, wants to require('pileup').
 # The dist/test files are required for code coverage
-babel src --ignore src/lib --out-dir dist
+babel src --retain-lines --ignore src/lib --out-dir dist
 cp -r src/lib dist/
 
 # Create dist/tests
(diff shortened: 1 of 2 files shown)


armish:
LGTM: I'm glad they exposed that argument and you didn't have to fork it ;)

Issue #313, opened 2015-10-16
Users should be able to pass arbitrary options to the tracks


armish:
We currently hardcode the decision to show specific visual components within the track. Ideally, users should be able to pass options to override these settings. For example:

- [x] connect paired reads (#110)
- [ ] show mismatches in the coverage (#233)
- [ ] show center line in the pileup track (#311)

danvk:
@armish was there a particular reason you wanted to optionalize showing mismatches & the center line?

Desktop IGV has an option for the coverage mismatch AF threshold, but not an explicit option to turn off mismatch rendering. So far as I can tell it doesn't have an option to turn off the center line.

I guess I'm just saying... these features are so great, can you think of a scenario where you'd want to turn them off? :)

armish:
yeah, agree that it doesn't make sense to turn the visualization options off at this point; but I feel bad about setting a hard threshold for the variant frequency, so maybe we can at least do that optional? :)

danvk:
IGV does have an option to show the center line:
http://www.broadinstitute.org/igv/Preferences#Alignments

Issue #254, opened 2015-09-18
Add a UI for toggling read pairs


danvk:
Currently pileup always shows paired reads. This behavior should be optional.

In IGV, this is done right right-click on the pileup track:

![screen shot 2015-09-18 at 2 32 24 pm](https://cloud.githubusercontent.com/assets/98301/9967687/1aa528de-5e12-11e5-9f1a-72c309f5b8c4.png)

In BioDalliance, this can only be done via the API.

How should pileup handle this? Ideas:

- Expose an API for updating visualization options and push the UI onto users of pileup (e.g. cycledash).
- Let each track handle its own UI for updating
- Have some common "track settings" code paths/UI elements

@jaclynperrone this would be a good thing to talk about with you!

PR #333, opened 2015-10-26
Add a viewAsPairs option

danvk:
See #254 
See #313

Syntax for passing options looks like this:

```javascript
  {
    viz: pileup.viz.pileup({
      viewAsPairs: true
    }),
    data: bamSource,
    cssClass: 'tumor',
    name: 'Alignments'
  }
```

I also took the opportunity to rename `playground-complete.html`. The name doesn't make much sense, since there's no other playground that's incomplete!

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/333)


danvk:
cc @ihodes would appreciate a look-over from you as well, since this is sort of a new user-facing API.

src/main/PileupCache.js
```diff
@@ -49,17 +43,29 @@ class PileupCache {
...
     this.refToPileup = {};
     this.referenceSource = referenceSource;
+    this.viewAsPairs = viewAsPairs;
+  }
+
+  // read name would make a good key, but paired reads from different contigs
+  // shouldn't be grouped visually. Hence we use read name + contig.
+  groupKey(read: Alignment): string {
+    if (this.viewAsPairs) {
+      return `${read.name}:${read.ref}`;
```
ihodes: is ref == contig?

src/main/PileupCache.js
```diff
@@ -91,13 +97,17 @@ class PileupCache {
...
       // TODO: if the mate Alignment is also available, it would be better to use that.
-      var intervals = [range];
-      var mateProps = read.getMateProperties();
-      if (mateProps && mateProps.ref && mateProps.ref == read.ref) {
-        mateInterval = new ContigInterval(mateProps.ref, mateProps.pos, mateProps.pos + range.length());
-        intervals.push(mateInterval);
+      if (this.viewAsPairs) {
+        var mateProps = read.getMateProperties();
+        var intervals = [range];
+        if (mateProps && mateProps.ref && mateProps.ref == read.ref) {
```
ihodes: Not at all necessary, just a thought; I've enjoyed using functions like `equalsIn(obj, path, val)` to handle this (in this case, `equalsIn(matProps, ['ref'], read.ref)`).

src/main/PileupCache.js
```diff
@@ -160,7 +170,7 @@ function spanAndInsert(intervals: ContigInterval<string>[]) {
   if (intervals.length == 1) {
     return {insert: null, span: intervals[0]};
   } else if (intervals.length !=2) {
-    throw `Called spanAndInsert with ${intervals.length} \in [1, 2]`;
+    throw `Called spanAndInsert with ${intervals.length} \notin [1, 2]`;
```
ihodes: Don't you wish the console would render LaTeX?

armish:
one speculative comment to get a conversation going about alternative ways to handle options, but this looks great!


---

Reviewed 12 of 12 files at r1.

---

<sup>**[src/main/pileup.js, line 135 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/333#-K1a42AYDvKXaMr24atP)** ([raw file](https://github.com/hammerlab/pileup.js/blob/b4247631cee9040c16da707b6944b826053e6b6d/src/main/pileup.js#L135)):</sup>
when I proposed #313, I was thinking about embedding default options to the tracks and make pileup pass an `opts` object to all of the components by default (whether empty or not); e.g.:

```
{
    viz: pileup.viz.pileup(),
    data: bamSource,
    cssClass: 'tumor',
    name: 'Alignments',
    opts: {viewAsPairs: true}
  },
{
    viz: pileup.viz.coverage(),
    data: bamSource,
    cssClass: 'tumor',
    name: 'Coverage',
    opts: {showMismatches: true}
  }
```
where tracks are responsible for handling whatever options are passed to them (e.g. overriding default ones with the incoming ones)

Looks like yours is a slightly different approach where instead of tracks, it is the pileup that knows about the default options, if I understand this correctly. I like this as it handles options in a generic way across tracks and doesn't lead to repetitive `_.extend` calls within tracks; but I don't like the fact that default options are hidden somewhere else in the code, which might make it harder to read the code for individual tracks and understand what each option does.

I don't really have strong opinions against either way, but love to hear what you guys think about the alternative?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/333)

ihodes:
@danvk LGTM, I like the option-passing (taking advantage of that decision early on, now).

danvk:
I've addressed all comments. Let me know what you guys think.


---

Review status: 9 of 11 files reviewed at latest revision, 3 unresolved discussions, some commit checks pending.

---

<sup>**[src/main/pileup.js, line 135 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/333#-K1a42AYDvKXaMr24atP)** ([raw file](https://github.com/hammerlab/pileup.js/blob/b4247631cee9040c16da707b6944b826053e6b6d/src/main/pileup.js#L135)):</sup>
You're totally right that tracks should set their own default options. I've added a `defaultOptions` property to `PileupTrack`.

I think it does make sense to pass option separately to the data source and the visualization (there will naturally be different options which are relevant to each). How to get two pileup tracks to share visualization options is an interesting question.


---

<sup>**[src/main/PileupCache.js, line 59 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/333#-K1_zq3bRSSXNuEiANtp-r1-59)** ([raw file](https://github.com/hammerlab/pileup.js/blob/b4247631cee9040c16da707b6944b826053e6b6d/src/main/PileupCache.js#L59)):</sup>
Yep, this is what Samtools calls it.


---

<sup>**[src/main/PileupCache.js, line 103 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/333#-K1_zq3bRSSXNuEiANtp-r1-103)** ([raw file](https://github.com/hammerlab/pileup.js/blob/b4247631cee9040c16da707b6944b826053e6b6d/src/main/PileupCache.js#L103)):</sup>
Maybe I'm too used to static languages, but `mateProps.ref` and `mateProps['ref']` feel like different things to me. (I realize they're exactly the same in JS).

Actually, there is a reason to prefer writing it this way. Flow can analyze it one way but not the other (e.g. it would catch if I wrote `contig` instead of `ref`).


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/333)

armish:
Awesome! :shipit: 


---

Reviewed 2 of 2 files at r2.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/333)

PR #338, opened 2015-10-27
Visual improvements to CoverageTrack

### LINKED ISSUE #334: Eliminate moiré pattern in Coverage track
Note how the white lines between coverage bins are more distinct in one location than the other:
coverage banding pattern

The solution is to use one `.fill()` call rather than repeated `.fillRect()` calls:
http://plnkr.co/edit/9QjNCm3X2jq7BgADjznt?p=preview

danvk:
This includes a bunch of changes:

1. eliminates the Moiré pattern (fixes #334). This required splitting out the mismatch rendering code into a separate loop.
1. rounds the y-axis coordinates to get crisper rendering. IGV.js does this and it looks much nicer.
1. fixes a bug where the bars were all ~10px too high, thus giving incorrect counts. Funny that we never noticed this!
1. restored the white border on the coverage track scale. I believe this was lost in the SVG→canvas transition (we used to do it with CSS).

Here's a before/after:

screen shot 2015-10-27 at 1 20 59 pm

screen shot 2015-10-27 at 1 20 14 pm

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/338)

### DIFF
--- a/src/main/CoverageTrack.js
+++ b/src/main/CoverageTrack.js
@@ -162,52 +162,66 @@ class CoverageTrack extends React.Component {
     ctx.reset();
     ctx.clearRect(0, 0, ctx.canvas.width, ctx.canvas.height);
 
-    var barWidth = xScale(1) - xScale(0),
-        barPadding = barWidth * style.COVERAGE_BIN_PADDING_CONSTANT;
+    var barWidth = xScale(1) - xScale(0);
+
+    var binPos = function(bin) {
+      var barPosX = xScale(bin.position),
+          barPosY = Math.round(yScale(bin.count) - yScale(axisMax));
+      return {barPosX, barPosY};
+    };
 
     // Draw coverage bins
-    this.binsInRange().forEach(bin => {
+    var bins = this.binsInRange();
+    var mismatchBins = [];  // keep track of which ones have mismatches
+    var vBasePosY = yScale(0);  // the very bottom of the canvas
+    ctx.fillStyle = style.COVERAGE_BIN_COLOR;
+    ctx.beginPath();
+    var {barPosX} = binPos(bins[0]);
+    ctx.moveTo(barPosX, vBasePosY);
+    bins.forEach(bin => {
       ctx.pushObject(bin);
-      var barPosX = xScale(bin.position),
-          barPosY = yScale(bin.count) - yScale(axisMax),
-          barHeight = Math.max(0, yScale(axisMax - bin.count));
-      // These are generic coverage bins
-      ctx.fillStyle = style.COVERAGE_BIN_COLOR;
-      ctx.fillRect(barPosX + barPadding,
-                   barPosY,
-                   barWidth - barPadding,
-                   barHeight);
+      var {barPosX, barPosY} = binPos(bin);
+      ctx.lineTo(barPosX, barPosY);
+      ctx.lineTo(barPosX + barWidth, barPosY);
 
-      // These are variant bars
       if (SHOW_MISMATCHES && !_.isEmpty(bin.mismatches)) {
-        var vBasePosY = yScale(0);  // the very bottom of the canvas
-        _.chain(bin.mismatches)
-          .map((count, base) => ({count, base}))  // pull base into the object
-          .sortBy(mc => -mc.count)  // the most common mismatch at the bottom
-          .each(mc => {
-            var {count, base} = mc;
-            if (count <= MISMATCH_THRESHOLD) {
-              // Don't show this as it doesn't have enough evidence
-              return;
-            }
-
-            var misMatchObj = {position: bin.position, count, base};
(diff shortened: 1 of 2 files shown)


armish:
Thanks for all these fixes, @danvk! Sorry about the 10px-bug, probably introduced it while transitioning to canvas :\

LGTM!


---

Reviewed 2 of 2 files at r1.

---

<sup>**[src/main/CoverageTrack.js, line 174 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/338#-K1eYxzEdqaB2ozcA14i)** ([raw file](https://github.com/hammerlab/pileup.js/blob/a15c96995339285e5cab206b984f2d9c4cdfb29e/src/main/CoverageTrack.js#L174)):</sup>
why do you have those parentheses around this statement?


---

<sup>**[src/main/CoverageTrack.js, line 185 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/338#-K1eZ5dDWVtE3k2eq93-)** ([raw file](https://github.com/hammerlab/pileup.js/blob/a15c96995339285e5cab206b984f2d9c4cdfb29e/src/main/CoverageTrack.js#L185)):</sup>
nice touch -- thanks!


---

<sup>**[src/main/CoverageTrack.js, line 196 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/338#-K1eZmXi36pOet-7w7v2)** ([raw file](https://github.com/hammerlab/pileup.js/blob/a15c96995339285e5cab206b984f2d9c4cdfb29e/src/main/CoverageTrack.js#L196)):</sup>
do you think it is worth keeping this padding only when the view is zoomed in at a level that you can distinguish individual bases? I don't like the Moiré pattern, but it looks better when you are investigating individual bases. 

Up to you.


---

<sup>**[src/main/CoverageTrack.js, line 226 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/338#-K1eZAzkSSiTRzmTH0Zz)** ([raw file](https://github.com/hammerlab/pileup.js/blob/a15c96995339285e5cab206b984f2d9c4cdfb29e/src/main/CoverageTrack.js#L226)):</sup>
aha! I didn't know you could strokeText on canvas and that is why I dropped those halos around the text :\ (TIL)


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/338)

danvk:
I added the padding back. Will merge on green.


---

Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

<sup>**[src/main/CoverageTrack.js, line 174 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/338#-K1eYxzEdqaB2ozcA14i)** ([raw file](https://github.com/hammerlab/pileup.js/blob/a15c96995339285e5cab206b984f2d9c4cdfb29e/src/main/CoverageTrack.js#L174)):</sup>
See the note on [this MDN page][1]. It's not valid JS syntax without the parens.

This is sort of a long-standing wart in JS... for example, if you type `{}` and hit enter in the Chrome console, it'll spit back `undefined`. I think it's trying to interpret `{}` as a block of code, not an object.

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#Syntax


---

<sup>**[src/main/CoverageTrack.js, line 196 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/338#-K1eZmXi36pOet-7w7v2)** ([raw file](https://github.com/hammerlab/pileup.js/blob/a15c96995339285e5cab206b984f2d9c4cdfb29e/src/main/CoverageTrack.js#L196)):</sup>
Yeah, it definitely looks better with the padding. I added it back & refactored this method a bit.

Here's what it looks like at different resolutions. I rounded all coordinates to get crisp lines. 
screen shot 2015-10-27 at 4 15 24 pm
screen shot 2015-10-27 at 4 21 32 pm
screen shot 2015-10-27 at 4 21 39 pm


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/338)

armish:
This looks even better! :)


---

Reviewed 2 of 2 files at r2.

---

<sup>**[src/main/CoverageTrack.js, line 174 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/338#-K1eYxzEdqaB2ozcA14i)** ([raw file](https://github.com/hammerlab/pileup.js/blob/a15c96995339285e5cab206b984f2d9c4cdfb29e/src/main/CoverageTrack.js#L174)):</sup>
aha, tricky one! Thanks for the link.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/338)

PR #343, opened 2015-10-29
Show counts for all bases when there are many mismatches

### LINKED ISSUE #337: Show all bases for mismatch bars in CoverageTrack
IGV shows colored bars for both ref & alt in the presence of variants:
screen shot 2015-10-27 at 1 11 04 pm

I like this. It makes it much clearer what's going on than our current rendering:

screen shot 2015-10-27 at 1 12 17 pm

danvk:
Fixes #337

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/343)


danvk:
This matches what I think is a reasonable behavior of IGV's.

@armish let me know what you think of this approach & I'll clean it up, factor out constants, etc.

armish:
Just to re-iterate our offline conversation on this: I have mixed feelings about showing the stack representing the reference base as it makes it harder to tell between the variant and the reference, but this style makes it variant locations much more prominent when you are zoomed out as you have more color/contrast within the coverage track. 

But let's go with this and match IGV's behavior. I like the hacky calculation, but we might hit edge cases where that heuristic might fail us (see my comment below)


---

Reviewed 5 of 5 files at r1.
Review status: all files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

<sup>**[src/main/CoverageTrack.js, line 87 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/343#-K1pI65ncCh3zDjaw3gK)** ([raw file](https://github.com/hammerlab/pileup.js/blob/b46a3c42f1f401295c182c872dccf0dd6b3e7915/src/main/CoverageTrack.js#L87)):</sup>
what if you have

A (ref) = %40
C (alt) = %30
T (alt) = %30

then this would flip a coin and pick either `C` or `T` and bump the reference up to %70, right?

You might want to do that %20 after the `forEach` logic to make sure that there is enough variants to make the ref calculation worth.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/343)

danvk:
I've updated tests, cleaned this up and addressed your comment below. Will submit on green.


---

Review status: 2 of 6 files reviewed at latest revision, 1 unresolved discussion.

---

<sup>**[src/main/CoverageTrack.js, line 87 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/343#-K1pI65ncCh3zDjaw3gK)** ([raw file](https://github.com/hammerlab/pileup.js/blob/b46a3c42f1f401295c182c872dccf0dd6b3e7915/src/main/CoverageTrack.js#L87)):</sup>
Not quite. In the case you gave, assuming depth 100, we'd have `count=100` and `mismatches = {C: 30, T: 30}`. So this would calculate `mismatchCount=60`. 60 is more than 20% of 100, so it would add in `mismatches.A = 100 - 60 = 40`.

The `forEach` is over loci, not over the mismatches at each locus.

In any case, this logic is kind of cryptic so I added some comments.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/343)

armish:
:+1:


---

Reviewed 4 of 4 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

<sup>**[src/main/CoverageTrack.js, line 87 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/343#-K1pI65ncCh3zDjaw3gK)** ([raw file](https://github.com/hammerlab/pileup.js/blob/b46a3c42f1f401295c182c872dccf0dd6b3e7915/src/main/CoverageTrack.js#L87)):</sup>
Ah, my bad -- looks like I misread the logic :\


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/343)

PR #346, opened 2015-10-30
Tiled Canvas

### LINKED ISSUE #86: Minimize redrawing of pileups
This is going to require some refactoring. Currently `PileupTrack` redoes the pileup & updates all the SVG elements on every frame. Ideally it would draw a superset of the visible reads whenever new data comes in. Panning would simply involve translating the whole SVG element.

danvk:
Fixes #86 

This introduces `TiledCanvas`, which maintains a cache of previously rendered buffers. This greatly improves performance during panning, since each frame just involves drawing pre-rendered buffers on the on-screen canvas. Panning at any zoom level improves to ~60fps after this change.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/346)

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -19,6 +19,8 @@
 .*node_modules/babel.*
 .*node_modules/d3.*
 .*node_modules/smash.*
+.*node_modules/source-map.*
+.*node_modules/exorcist.*
 .*react/node_modules/fbjs.*
 .*coverage.*
 
--- a/README.md
+++ b/README.md
@@ -74,18 +74,6 @@ To play with the demo, start an [http-server][hs]:
 
 Then open [http://localhost:8080/examples/index.html](http://localhost:8080/examples/index.html) in your browser of choice.
 
-### BioJS
-
-Alternatively you can use [BioJS sniper][sniper] for the demo:
-
-    npm install -g sniper # installs sniper globally
-
-Once installed, start sniper in the `pileup.js` folder:
-
-    sniper # and keep it running
-
-And browse to [http://localhost:9090/examples/playground](http://localhost:9090/examples/playground).
-
 ## Testing
 
 Run the tests from the command line:
@@ -97,15 +85,33 @@ Run the tests in a real browser:
     npm run http-server
     open http://localhost:8080/src/test/runner.html
 
-To continuously regenerate the combined JS, run:
+To continuously regenerate the combined pileup and test JS, run:
 
     npm run watch
+
+To run a single test from the command line, use:
+
+    npm run test -- --grep=pileuputils
+
+To do the same in the web UI, pass in a `?grep=` URL parameter.
 
 To typecheck the code, run
 
     npm run flow
 
 For best results, use one of the flowtype editor integrations.
+
+### BioJS
+
+Alternatively you can use [BioJS sniper][sniper] for the demo:
+
+    npm install -g sniper # installs sniper globally
+
+Once installed, start sniper in the `pileup.js` folder:
+
+    sniper # and keep it running
+
+And browse to [http://localhost:9090/examples/playground](http://localhost:9090/examples/playground).
 
 ## License
 
--- a/lib/underscore.js
+++ b/lib/underscore.js
@@ -45,4 +45,7 @@ declare module "underscore" {
 
   declare function reduce<T, R>(a: T[], fn: (memo: R, val: T, idx: number, list: T[])=>R, memo?: R): R;
   declare function reduce<K, T, R>(a: {[key:K]: T}, fn: (memo: R, val: T, key: K, o: {[key:K]: T})=>R, memo?: R): R;
+
+  declare function uniq<T>(array: T[], isSorted?: boolean, iteratee?: (v: T)=>any): T[];
+  declare function unique<T>(array: T[], isSorted?: boolean, iteratee?: (v: T)=>any): T[];
 }
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
   "dependencies": {
     "backbone": "1.1.2",
     "d3": "^3.5.5",
-    "data-canvas": "0.0.0",
+    "data-canvas": "^0.1.0",
     "jbinary": "^2.1.3",
     "jdataview": "^2.5.0",
     "pako": "^0.2.5",
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -18,4 +18,4 @@ trap finish EXIT
 sleep 1
 
 # Start the tests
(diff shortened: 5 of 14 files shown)


ihodes:
This is great. What about testing 1) that it (TiledCanvas) renders the right thing when it's rendering (not cached) and 2) the cache serves the same thing (pixelwise) that TiledCanvas would render were the tile not cached?

danvk:
If I understand correctly, you're suggesting testing the renderPileup
function directly, and then testing the TiledCanvas separately? That would
actually be great. We'd be testing the pileup rendering independently from
React.

armish:
Really nice! I am still trying to wrap my mind around it, but re: tests: option 2 sounds like the way to go, despite all the complications regarding the offset math that will come with it.

Also: how does this interface with capturing clicks? Will you be re-rendering only the tile (is it possible?) or everything (as usual)?


---

Reviewed 11 of 11 files at r1.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

<sup>**[scripts/test.sh, line 21 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/346#-K28uRnfNfsTaR2ASpME)** ([raw file](https://github.com/hammerlab/pileup.js/blob/32afd887421321c4c53247d3ec14f7aecedffaa3/scripts/test.sh#L21)):</sup>
any hints/tricks on useful arguments to be passed to `mocha-phantomjs`? :)


---

<sup>**[src/main/PileupTrack.js, line 43 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/346#-K28yNMcBjbFuH6we7Po)** ([raw file](https://github.com/hammerlab/pileup.js/blob/32afd887421321c4c53247d3ec14f7aecedffaa3/src/main/PileupTrack.js#L43)):</sup>
is this going to be the only thing that tracks need to implement to leverage this tile functionality? 


---

<sup>**[src/main/PileupTrack.js, line 51 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/346#-K28ym9lGnNtYRacFSEn)** ([raw file](https://github.com/hammerlab/pileup.js/blob/32afd887421321c4c53247d3ec14f7aecedffaa3/src/main/PileupTrack.js#L51)):</sup>
If I am understanding this correctly, this is going to be equal to the fixed height for all the trivial tracks that don't support scrolling, right? e.g. for LocationTrack, this is going to be 50.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/346)

danvk:
I wound up going with option 2 (make `drawImage` forward recorded calls). This works well, though we do have to be careful to avoid duplicate objects. For example, if an alignment intersects two tiles, then its objects will get `push`ed and `pop`ped multiple times. I tweaked `PileupTrack` and some of the tests to accommodate this.


---

Review status: 7 of 14 files reviewed at latest revision, 3 unresolved discussions.

---

<sup>**[scripts/test.sh, line 21 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/346#-K28uRnfNfsTaR2ASpME)** ([raw file](https://github.com/hammerlab/pileup.js/blob/32afd887421321c4c53247d3ec14f7aecedffaa3/scripts/test.sh#L21)):</sup>
The only one I know of is `--grep`, e.g.

```
npm run test -- --grep=PileupTrack
```

I added a note about this to `README.md`.


---

<sup>**[src/main/PileupTrack.js, line 43 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/346#-K28yNMcBjbFuH6we7Po)** ([raw file](https://github.com/hammerlab/pileup.js/blob/32afd887421321c4c53247d3ec14f7aecedffaa3/src/main/PileupTrack.js#L43)):</sup>
I think so! Pretty easy, right?


---

<sup>**[src/main/PileupTrack.js, line 51 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/346#-K28ym9lGnNtYRacFSEn)** ([raw file](https://github.com/hammerlab/pileup.js/blob/32afd887421321c4c53247d3ec14f7aecedffaa3/src/main/PileupTrack.js#L51)):</sup>
That's correct. I don't think `LocationTrack` is a good fit for `TiledCanvas` though, since it changes its content in ways that aren't simple pans (i.e. it changes the location text depending on where you are). `CoverageTrack`, `GeneTrack` and `GenomeTrack` are good candidates for porting.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/346)

armish:
Was just going to ask you about that test and uniqueness, but now it makes sense. Good to know about this behavior!

:+1:


---

Reviewed 7 of 9 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

<sup>**[src/main/PileupTrack.js, line 43 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/346#-K28yNMcBjbFuH6we7Po)** ([raw file](https://github.com/hammerlab/pileup.js/blob/32afd887421321c4c53247d3ec14f7aecedffaa3/src/main/PileupTrack.js#L43)):</sup>
yeah, I like it!


---

<sup>**[src/main/PileupTrack.js, line 388 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/346#-K2EC41ZNElkRCAwL_Re)** ([raw file](https://github.com/hammerlab/pileup.js/blob/d58734e2822fa9380d08104018692e133489cd9a/src/main/PileupTrack.js#L388)):</sup>
you don't seem to have tests for this, right now. Do you? Missing file?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/346)

danvk:
Oh, and to answer your question about click tracking: I do re-render the whole scene (as before), without tiling.


---

Review status: 12 of 14 files reviewed at latest revision, 1 unresolved discussion.

---

<sup>**[src/main/PileupTrack.js, line 388 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/346#-K2EC41ZNElkRCAwL_Re)** ([raw file](https://github.com/hammerlab/pileup.js/blob/d58734e2822fa9380d08104018692e133489cd9a/src/main/PileupTrack.js#L388)):</sup>
I don't but I should. It's a to-do.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/346)

armish:
Thanks :ship:


---

Reviewed 2 of 2 files at r3.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/346)

Issue #351, opened 2015-11-03
TiledCanvas follow-up work


danvk:
- [ ] Migrate GenomeTrack
- [x] Migrate CoverageTrack
- [x] Make sure mismatched base pairs don't get lost on tile boundaries
- [x] Make sure alignment arrows don't get lost on tile boundaries

PR #353, opened 2015-11-04
Tile the CoverageTrack

danvk:
See #351

Adding explicit types for props/state works really nicely. It makes Flow much more useful for these components.

The hack for PhantomJS is kinda gross, but I don't see an alternative.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/353)

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,7 +53,7 @@
   "dependencies": {
     "backbone": "1.1.2",
     "d3": "^3.5.5",
-    "data-canvas": "^0.1.0",
+    "data-canvas": ">=0.1.1",
     "jbinary": "^2.1.3",
     "jdataview": "^2.5.0",
     "pako": "^0.2.5",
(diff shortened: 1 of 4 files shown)


danvk:
Update on this:

- It's really nice—panning is super fast once the tiles are loaded
- I belive it's failing because PhantomJS doesn't support flexbox :(

armish:
Thanks for these changes -- especially for adding those flow-types, Dan! It does feel super fast  :dash:!


---

Reviewed 3 of 4 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

<sup>**[src/main/CoverageTrack.js, line 124 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/353#-K2NRzA85BsesFxXwOqr)** ([raw file](https://github.com/hammerlab/pileup.js/blob/16e7dac9b1232de43e725168f6a1da74dacf84a1/src/main/CoverageTrack.js#L124)):</sup>
was expecting a much worse solution than this :)


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/353)

PR #354, opened 2015-11-05
Props states types

### LINKED ISSUE #235: Add types for props and state
It should be possible to specify types for `props` and `state` using Flow:
https://github.com/facebook/flow/blob/28f01d1800618ad9f679d7a394c1feff1edf3a16/lib/react.js#L16

This would greatly improve type safety for visualization tracks.

danvk:
Fixes #235 

This migrates all React components to use ES6 classes and declares types for `state` and `props` in all of them. This should make Flow much more effective.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/354)

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -22,6 +22,7 @@
 .*node_modules/source-map.*
 .*node_modules/exorcist.*
 .*react/node_modules/fbjs.*
+.*/node_modules/react/.*
 .*coverage.*
 
 [include]
--- a/package.json
+++ b/package.json
@@ -60,7 +60,6 @@
     "q": "^1.1.2",
     "react": "^0.14.0",
     "react-dom": "^0.14.0",
-    "react-addons-pure-render-mixin": "^0.14.0",
     "shallow-equals": "0.0.0",
     "underscore": "^1.7.0"
   },
--- a/src/main/Alignment.js
+++ b/src/main/Alignment.js
@@ -41,6 +41,6 @@ export type Alignment = {
 export type AlignmentDataSource = {
   rangeChanged: (newRange: GenomeRange) => void;
   getAlignmentsInRange: (range: ContigInterval<string>) => Alignment[];
-  on: (event: string, handler: Function) => void;
+  on: (event: string, handler: Function) => void;  // really => AlignmentDataSource
   off: (event: string) => void;
 };
--- a/src/main/BigBedDataSource.js
+++ b/src/main/BigBedDataSource.js
@@ -24,7 +24,7 @@ export type Gene = {
 }
 
 // Flow type for export.
-type BigBedSource = {
+export type BigBedSource = {
   rangeChanged: (newRange: GenomeRange) => void;
   getGenesInRange: (range: ContigInterval<string>) => Gene[];
   on: (event: string, handler: Function) => void;
--- a/src/main/Controls.js
+++ b/src/main/Controls.js
@@ -18,7 +18,7 @@ type Props = {
   onChange: (newRange: GenomeRange)=>void;
 };
 
-class Controls extends (React.Component : typeof ReactComponent) {
+class Controls extends React.Component {
   props: Props;
   state: void;  // no state
 
--- a/src/main/Controls.js
+++ b/src/main/Controls.js
@@ -9,8 +9,7 @@ import type {PartialGenomeRange} from './types';
 var React = require('react'),
     _ = require('underscore');
 
-var types = require('./react-types'),
-    utils = require('./utils'),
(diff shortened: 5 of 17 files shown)


danvk:
It's interesting that the bundle expands as a result of this. I wonder if this means that babel-ified ES6 classes are significantly more verbose than `React.createClass`?

armish:
I'm learning quite a lot about flow and ES6 features just by reviewing these changes. Looks great! :100:

BTW: How much of an increase in size are we talking about? Couldn't see that from the PR status:

Imgur


---

Reviewed 5 of 14 files at r1, 13 of 13 files at r2.
Review status: all files reviewed at latest revision, 3 unresolved discussions, some commit checks failed.

---

<sup>**[.flowconfig, line 25 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/354#-K2PFm1_GijtjwgzAQRK)** ([raw file](https://github.com/hammerlab/pileup.js/blob/5119a848ab52821cc38505c0db10e7f682089143/.flowconfig#L25)):</sup>
just curious: at what point you decide to start ignoring a dependency for type checks?


---

<sup>**[src/main/Controls.js, line 23 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/354#-K2PG72zO8luNoiSpcE8)** ([raw file](https://github.com/hammerlab/pileup.js/blob/5119a848ab52821cc38505c0db10e7f682089143/src/main/Controls.js#L23)):</sup>
this is nice. Will we be getting type errors if we try to set the state by mistake then?


---

<sup>**[src/main/Controls.js, line 108 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/354#-K2PGMOHGV41lfR4XTWc)** ([raw file](https://github.com/hammerlab/pileup.js/blob/5119a848ab52821cc38505c0db10e7f682089143/src/main/Controls.js#L108)):</sup>
what is the difference between `onSubmit={this.handleFormSubmit}` and `{this.handleFormSubmit.bind(this)}`?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/354)

danvk:
You and me both. I wound up filing an issue against Flow to figure out that I needed to `[ignore]` React.

The ∆ disappeared because of merge conflicts (which mean that we lost the "pr" build). It was something like +4k. Once I rebase it should reappear.


---

Review status: 16 of 17 files reviewed at latest revision, 3 unresolved discussions.

---

<sup>**[.flowconfig, line 25 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/354#-K2PFm1_GijtjwgzAQRK)** ([raw file](https://github.com/hammerlab/pileup.js/blob/5119a848ab52821cc38505c0db10e7f682089143/.flowconfig#L25)):</sup>
There are two reasons to ignore a directory:

1. We have type definitions for the library it contains. Unless you explicitly ignore `node_modules/underscore`, then the type declarations file for underscore won't be used. This is surprising and pretty dumb behavior. The Flow team is supposed to fix it soon.
1. It's a dev dependency, rather than one that we `require`, and hence not something that Flow should care about.

It would be nice if we could just ignore all of `node_modules`, but then `require`ing libraries for which we don't have type definitions would confuse Flow. Maybe we could get away with it some day when we have type definitions for most libraries.


---

<sup>**[src/main/Controls.js, line 23 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/354#-K2PG72zO8luNoiSpcE8)** ([raw file](https://github.com/hammerlab/pileup.js/blob/5119a848ab52821cc38505c0db10e7f682089143/src/main/Controls.js#L23)):</sup>
Yep :)


---

<sup>**[src/main/Controls.js, line 108 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/354#-K2PGMOHGV41lfR4XTWc)** ([raw file](https://github.com/hammerlab/pileup.js/blob/5119a848ab52821cc38505c0db10e7f682089143/src/main/Controls.js#L108)):</sup>
It's subtle. In old-style classes, React would "autobind" your handlers, so that `this` was set to the component inside them. With ES6 classes this is disabled, so we have to do it ourselves.

Check out this [blog post][1].

[1]: https://facebook.github.io/react/blog/2015/01/27/react-v0.13.0-beta-1.html#autobinding


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/354)

danvk:
screen shot 2015-11-06 at 11 04 44 am

Issue #181, opened 2015-06-11
Should reads containing variants be sorted to the top?


jaclynperrone:
An open question. It looks like IGV is doing it, which makes the pileup view a little more coherent. Thoughts?

danvk:
Here's what the [IGV manual has to say][1]:

> Users can also specifiy color and also sort reads by various options, including start location, strand, nucleotide, mapping quality, sample tag, or read group tag. For a description of all user-specified color and sort options, see the alignment track pop-up menu.
> 
> For example, to sort alignments:
>
> Right-click a track for the pop-up menu.
> Select a Sort option from the pop-up menu. IGV sorts the alignments that intersect the center line of the display.
> Sorting rearranges rows so that alignments that intersect the center of the display appear in the order specified.  This can cause the alignment layout away from the center line to appear sparse.   To restore the layout to an optimally packed configuration, select Re-pack alignments from the pop-up menu.
>
> Repeat the most recent sort with hotkey ctrl-s.

My interpretation is that the default behavior is to first pile up the reads which intersect the center line, then pile up the others.

[1]: http://www.broadinstitute.org/igv/AlignmentData#thmr_2

PR #355, opened 2015-11-05
Implement PileupCache.sortReadsAt

danvk:
See #181 

There's no UI for this yet. I'm thinking of two ideas there:

1. When you go to a new location (either on the initial load or by jumping to a previously unloaded region), then the alignments should get sorted around the center position.
1. There could be an option to sort the reads via the UI, maybe under a gear icon below the pileup track label.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/355)

### DIFF
--- a/lib/underscore.js
+++ b/lib/underscore.js
@@ -16,13 +16,16 @@ declare module "underscore" {
   declare function chain<S>(obj: S): any;
   declare function any<T>(list: Array<T>, pred: (el: T)=>boolean): boolean;
 
-  declare function each<T>(o: {[key:string]: T}, iteratee: (val: T, key: string)=>void): void;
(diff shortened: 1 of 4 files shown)


armish:
If possible, I would suggest thinking about regions of interest instead of a particular position in the genome, where sorting reads at a single base would mean the region of interest is only 1bp long. It might be unnecessary to sort around the center position especially if the view is zoomed out, but would much make much more sense if you are to sort around center +/- X bps. Opt 2 also sound great and would make it much easier to investigate variants at a single position (maybe a right click context menu would also help here).

:+1:


---

Reviewed 4 of 4 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/355)

PR #362, opened 2015-11-11
Color read pairs with abnormal inserts

### LINKED ISSUE #256: Color read pairs
I've found myself using the IGV options to color read pairs by:

- [ ] insert size
- [ ] pair orientation

There are especially helpful for picking out reads which support a structural variant:

screen shot 2015-09-18 at 5 24 19 pm

(the red pairs have opposite-from-expected orientation)

danvk:
Fixes #256 

I've tried to match IGV's behavior as best I could. It highlights pairs which are in the extreme 0.5%iles on either end. This seems a little strange to me, though, since it would imply that the longest and shortest visible reads will always be marked as outliers. In any case, we can refine this in the future.

Other changes:
- Restructured examples
- Killed BioJS/Sniper (we think it was broken anyway)
- Added a BAM based on synth4 and a demo which uses it to highlight SVs

screen shot 2015-11-11 at 6 34 31 pm

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/362)


danvk:
FYI, the reason you don't see short inserts rendered in blue is that the min insert size gets set to zero. I need to investigate what's going on here.

danvk:
The issue was that what I was calling an "insert" was not what's usually thought of as an insert. The "insert" is the tip-to-tip distance for the read pair, not the distance between the inner ends of the mated reads. I adjusted the logic to reflect this. Now you can see blue (shorter than expected) reads in the screenshots.

I also added some constants for filtering extremely long inserts and suppressing coloring of abnormal inserts when there aren't enough reads to generate reliable statistics.

armish:
this looks really great, Dan! I am really excited to see this helping people identify potentially interesting reads :science:

I have a few basic questions regarding the behavior of the code, but feel to merge on green if they are false alarms :+1:


---

Reviewed 13 of 24 files at r1, 7 of 15 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

<sup>**[package.json, line 15 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/362#-K2wZo1az3IEC9FOP1Wk)** ([raw file](https://github.com/hammerlab/pileup.js/blob/d8562f7e07a7e88952313e82b6d73dcb5f7309a4/package.json#L15)):</sup>
you might want to remove this keyword, too. Since we are dropping the BioJS support, we probably don't want them to index us in their repository.


---

<sup>**[src/main/GA4GHAlignment.js, line 98 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/362#-K2w_ZBzgN07_JqVIXdH)** ([raw file](https://github.com/hammerlab/pileup.js/blob/d8562f7e07a7e88952313e82b6d73dcb5f7309a4/src/main/GA4GHAlignment.js#L98)):</sup>
do you need to wrap this with a `Math.abs`? Is there any chance that the directionality of the reads could mass with this calculation?


---

<sup>**[src/main/PileupCache.js, line 244 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/362#-K2wbhh9mLAavG6sEyXQ)** ([raw file](https://github.com/hammerlab/pileup.js/blob/d8562f7e07a7e88952313e82b6d73dcb5f7309a4/src/main/PileupCache.js#L244)):</sup>
sorry I am missing an obvious implementation detail here; but, will this **always** label some of the reads as outliers since a group of reads will always have a 5 and 95 percentile in them. I am asking this because if it is the case, then you will always have blue and red read pairs in the view even though they are probably normal reads. This of course will work great when you have real outliers, but will probably produce false positive outliers in some cases.

This is just nit-picking (and can be addressed by a different PR), but do you think it would make sense to implement a simple outlier detection mechanism instead of relying on percentiles for this?  (related: http://www.purplemath.com/modules/boxwhisk3.htm) 


---

<sup>**[src/main/PileupTrack.js, line 92 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/362#-K2wavxKqG0vwQThCd6F)** ([raw file](https://github.com/hammerlab/pileup.js/blob/d8562f7e07a7e88952313e82b6d73dcb5f7309a4/src/main/PileupTrack.js#L92)):</sup>
do you think it is worth adding a simple comment to this to explain why you need this (maybe by copy/paste'ing what you have as a PR comment)?


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/362)

danvk:
Thanks for looking this over. I've addressed your comments & will merge on green.


---

Review status: all files reviewed at latest revision, 4 unresolved discussions, some commit checks failed.

---

<sup>**[package.json, line 15 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/362#-K2wZo1az3IEC9FOP1Wk)** ([raw file](https://github.com/hammerlab/pileup.js/blob/d8562f7e07a7e88952313e82b6d73dcb5f7309a4/package.json#L15)):</sup>
Sure, removed.


---

<sup>**[src/main/GA4GHAlignment.js, line 98 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/362#-K2w_ZBzgN07_JqVIXdH)** ([raw file](https://github.com/hammerlab/pileup.js/blob/d8562f7e07a7e88952313e82b6d73dcb5f7309a4/src/main/GA4GHAlignment.js#L98)):</sup>
If an `Interval`'s `stop` is less than its `start`, then it's empty. That shouldn't happen.


---

<sup>**[src/main/PileupCache.js, line 244 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/362#-K2wbhh9mLAavG6sEyXQ)** ([raw file](https://github.com/hammerlab/pileup.js/blob/d8562f7e07a7e88952313e82b6d73dcb5f7309a4/src/main/PileupCache.js#L244)):</sup>
Yeah, see comment in PR description :)

> I've tried to match IGV's behavior as best I could. It highlights pairs which are in the extreme 0.5%iles on either end. This seems a little strange to me, though, since it would imply that the longest and shortest visible reads will always be marked as outliers. In any case, we can refine this in the future.

The minimum number of reads threshold should take care of really egregious cases. Let's play around with this for now and see what we think.


---

<sup>**[src/main/PileupTrack.js, line 92 \[r2\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/362#-K2wavxKqG0vwQThCd6F)** ([raw file](https://github.com/hammerlab/pileup.js/blob/d8562f7e07a7e88952313e82b6d73dcb5f7309a4/src/main/PileupTrack.js#L92)):</sup>
Done.


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/362)

armish:
Reviewed 2 of 2 files at r3.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/362)

Issue #22, opened 2015-03-09
Create a class for mapping ranges to objects


danvk:
It seems like this will be an incredibly common need for any tool that works with genomic data.

Examples:

- Mapping a range of loci to their nucleotides in the reference genome.
- A cache containing subsets of a remote file.

A class for this could do intelligent things like coalescing adjacent ranges, extracting subranges, etc.

danvk:
This is known as an [interval tree][1]. There's a [JavaScript implementation][2] on NPM which I could either use directly or adapt for use with my `Interval` class.

[1]: http://en.wikipedia.org/wiki/Interval_tree
[2]: https://www.npmjs.com/package/interval-tree

PR #367, opened 2015-11-12
Improvements to reference source

### LINKED ISSUE #97: Optimize TwoBitDataSource.getRange
It's showing up prominently in the flamecharts, using way more cycles than the seemingly more complicated `getAlignmentsInRange`:

screen shot 2015-04-23 at 5 26 58 pm

### LINKED ISSUE #363: Mismatches not shown initially at coarse zoom levels
When you open the `structural-variants.html` demo, it initially shows no mismatches. This isn't because there aren't any. Rather, it's because the reference wasn't loaded at this zoom level.

We should still load the reference when zoomed out, even if we don't show it. You can see the mismatches if you zoom in and then back out again.

danvk:
This is a big optimization to `TwoBitDataSource`. Rather than storing a giant hash table:

```json
{
  "chr17:12345": "A",
  "chr17:12346": "C",
  "chr17:12347": "T",
  ...
}
```

it instead stores 1000 character strings. This should result in much lower memory usage and hopefully it should be faster, since it requires ~1000x fewer hash lookups.

I took the opportunity to bump up the amount of reference data that is loaded by default, and the amount which is permissible to load. 2bit data is cheap. 100kbp is only ~25KB of data. Now that it's stored efficiently, there's little harm in fetching extra.

Fixes #97 
Fixes #363 
See #22 
See #351 

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/367)


armish:
Thorough job! Nothing really caught my eye :+1:


---

Reviewed 16 of 16 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

<sup>**[src/main/ContigInterval.js, line 91 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/367#-K2yf5UrsNXjo_vMXfXw)** ([raw file](https://github.com/hammerlab/pileup.js/blob/ced6079c3dc9acf26525e58572bf20d65c8e2872/src/main/ContigInterval.js#L91)):</sup>
is there a specific reason why you didn't go with `_.isString`? Just curious, because I saw that you added it to types.


---

<sup>**[src/main/GenomeTrack.js, line 79 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/367#-K2yhdPK34VkjB9sNWXG)** ([raw file](https://github.com/hammerlab/pileup.js/blob/ced6079c3dc9acf26525e58572bf20d65c8e2872/src/main/GenomeTrack.js#L79)):</sup>
that is a new syntactic sugar for me! Does it mean  of type `VizProps` but with a `TwoBitSource` `source` property? Cool!


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/367)

danvk:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

<sup>**[src/main/ContigInterval.js, line 91 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/367#-K2yf5UrsNXjo_vMXfXw)** ([raw file](https://github.com/hammerlab/pileup.js/blob/ced6079c3dc9acf26525e58572bf20d65c8e2872/src/main/ContigInterval.js#L91)):</sup>
I would have preferred to, but Flow doesn't understand that it refines the type of `this.contig`. Nor could it, really.


---

<sup>**[src/main/GenomeTrack.js, line 79 \[r1\]](https://reviewable.io:443/reviews/hammerlab/pileup.js/367#-K2yhdPK34VkjB9sNWXG)** ([raw file](https://github.com/hammerlab/pileup.js/blob/ced6079c3dc9acf26525e58572bf20d65c8e2872/src/main/GenomeTrack.js#L79)):</sup>
Yep, it's an [intersection type](http://flowtype.org/docs/union-intersection-types.html#)!


---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/367)

Issue #375, opened 2015-11-16
Only invalidate tiles which need to be updated


danvk:
`TiledCanvas` only has an `invalidateAll` method. In most cases, only a subset of the tiles should be invalidated when new data comes in. This should result in a speedup.

PR #376, opened 2015-11-16
Add TiledCanvas.invalidateRange

danvk:
This is an easy win. Getting it right for `CoverageTrack` is going to be a little trickier, since new data could change the y-axis range.

See #375 

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/376)


armish:
This looks good - I guess we can re-scale the images based on how y-axis changes, but could be much trickier than said.


---

Reviewed 3 of 3 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/376)

danvk:
Yeah, the solution there is to:

1. Invalidate a range if it `maxCoverage` hasn't changed.
2. Invalidate everything if it has.

PR #377, opened 2015-11-18
Remove white space around canvas items

armish:
Looks like [the mysterious-vertical-space bug](https://github.com/hammerlab/cycledash/pull/811) is at it again: http://www.hammerlab.org/pileup/

screen recording 2015-11-18 at 11 54 am

This removes the default white space around `canvas` elements and fixes the problem.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/377)


danvk:
Thanks for the fix, @armish

PR #381, opened 2015-11-18
Only fetch new ranges in {TwoBit,Bam}DataSource

### LINKED ISSUE #379: data sources should understand partial cache hits
Say you view `chr17:10,000-20,000`. This will eventually trigger `newdata` events on the bam & 2bit sources with a range of `[10,000, 20,000]`.

Then you pan left 1,000bp. This will request `chr17:9,000-19,000` and fire `newdata` events with ranges of `[9,000, 19,000]`.

This is pretty wasteful! It should only request `chr17:9,000-10,000` (i.e. the new portion of the region) and fire `newdata` events for that range as well.

danvk:
Fixes #379 

This is a _really_ nice performance win. See explanation in issue #379. After this, panning at 16,000bp resolution is pretty reasonable.

There were no screenshot diffs.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/381)

### DIFF
--- a/src/main/Alignment.js
+++ b/src/main/Alignment.js
@@ -43,5 +43,6 @@ export type AlignmentDataSource = {
   rangeChanged: (newRange: GenomeRange) => void;
   getAlignmentsInRange: (range: ContigInterval<string>) => Alignment[];
   on: (event: string, handler: Function) => void;  // really => AlignmentDataSource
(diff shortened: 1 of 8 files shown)


danvk:
cc @armish

ihodes:
Nice win, LGTM!

Issue #350, opened 2015-11-03
Center lines get out of sync


danvk:
If you pan around wildly in the demo, you'll eventually notice that the two center lines get out of sync by 1 base pair. If you keep panning around, they'll come back together.

danvk:
This seems to have something to do with one of the `div.drag-wrapper` elements getting shifted over.

danvk:
It also has something to do with vertical scrollbars on the pileup track. If I set `overflow-y: hidden`, I can't get the lines out of sync. It's quite easy to do otherwise.

PR #383, opened 2015-11-20
Force pileup tracks to stay aligned

danvk:
The issue is that one of the pileups winds up with `scrollLeft=15`. I'm not sure why. It only happens in Chrome.

You can still see the issue after this change while you're panning, but it goes away once you let go of the mouse.

Mitigates #350

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/383)


ihodes:
Crazy… LGTM!

PR #386, opened 2015-11-20
Show a Zenodo-based DOI badge to allow citations of specific releases

armish:
This will help people who would like cite specific versions of the pileup.js for their papers.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/386)


danvk:
👍

PR #388, opened 2015-11-23
Record version in pileup.version

danvk:
This will be helpful if we ever want to check what version of pileup someone is using, e.g. ourselves on hammerlab.org or someone else on dnanexus.

Nothing ever stays in sync unless you force it to, so I've added a lint check that the two version numbers are the same.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/388)

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -31,8 +31,12 @@ browserify \
 # Create dist/pileup.js.map
 cat dist/pileup.js | exorcist --base . dist/pileup.js.map > /dev/null
 
+version=$(grep '"version": ' package.json | sed 's/.*: "//; s/".*//')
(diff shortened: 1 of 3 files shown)


armish:
That's great - I like the lint check you put in place for this. I also wonder if we can make this part of the build process (something like an automatic header appended to the final file).

danvk:
I added a [jQuery-style][1] preamble to the minified JS. It looks like this:

```
/*! pileup v0.6.1 | (c) 2015 HammerLab | Apache-2.0 licensed */
```

[1]: https://cdnjs.cloudflare.com/ajax/libs/jquery/3.0.0-alpha1/jquery.min.js

PR #392, opened 2015-11-23
Don't stretch tiles vertically

### LINKED ISSUE #384: Tile shearing on pan + zoom out + pan
Pan a bit to the right on the demo page. Zoom out. Pan a bit more. The tiles get misaligned:

screen shot 2015-11-20 at 4 33 42 pm

I usually see this somewhere around `chr17:7,513,869`.

danvk:
Fixes #384 

The issue was that the track height would increase (because the pileup got taller due to added reads) and the existing tiles would be stretched vertically to match it. This was goofy. Tiles should be rendered at their natural height.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/392)

### DIFF
--- a/src/main/TiledCanvas.js
+++ b/src/main/TiledCanvas.js
@@ -95,7 +95,9 @@ class TiledCanvas {
       // rounding issues, which can result in 1px gaps or overdrawing.
       // We always have:
       //   width - tile.buffer.width \in {-1, 0, +1}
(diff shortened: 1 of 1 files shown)


armish:
LTGM :+1:

---

Reviewed 1 of 1 files at r1.

---



_Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/hammerlab/pileup.js/392#-:-K3qnampVJagxCa90S86:386398759)_

PR #400, opened 2015-12-18
Update to Flow 0.20

danvk:
See their [changelog][1]. The only really interesting thing here is the bug it uncovered in `GA4GHAlignment.js`.

[1]: https://github.com/facebook/flow/releases/tag/v0.20.0

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/400)

### DIFF
--- a/src/main/GA4GHAlignment.js
+++ b/src/main/GA4GHAlignment.js
@@ -94,7 +94,7 @@ class GA4GHAlignment /* implements Alignment */ {
       var start1 = this._interval.start(),
           stop1 = this._interval.stop(),
           start2 = m.pos,
-          stop2 = start2 + this.getSequence.length();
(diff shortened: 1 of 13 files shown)


src/main/GA4GHAlignment.js
```diff
@@ -94,7 +94,7 @@ class GA4GHAlignment /* implements Alignment */ {
       var start1 = this._interval.start(),
           stop1 = this._interval.stop(),
           start2 = m.pos,
-          stop2 = start2 + this.getSequence.length();
+          stop2 = start2 + this.getSequence().length;
```
danvk: !?!?!?

ihodes: That's a bit terrifying!

ihodes:
LGTM

PR #403, opened 2016-02-13
Adding the ability to access BAM files via CGI / samtools

selkovjr:
Sorry for jumping in with a possibly dumb idea. I run a massive study of variants where I need to document  small pileup regions (usually no larger than a dozen bases on either side). So far, nobody I know has figured a better way of capturing small pileups than by taking screenshots from Tablet or IGV. That gets tiresome really fast and updating the views with changes in the data is next to impossible.

This branch is my first experiment with pileup.js. I have basically achieved what I wanted, and it works great, as long as navigation is disabled. I think I misunderstood the read cache and it overflows while zooming or panning more than a couple steps from the initial view. When that happens, the pileup canvas fails to render (but it is mapped correctly, as I can tell by clicking on invisible reads).

Maybe I got the whole thing wrong, and I would love to hear from an expert about the proper way of doing it. But even the way it is, it solves the core problem -- I have pileups in web pages linked to data.

[Review on Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/403)


armish:
thanks for the PR, @selkovjr -- it is great to hear a really good use case of pileup.js, especially that integrates well with samtools. I was out of the office for the last few days, but will be taking a look at the code in the next few days. Is it OK for you to continue this review via Reviewable or do you prefer the standard inline GitHub comments? Let me know.

selkovjr:
Thanks @armish; I haven't seen Reviewable before -- it looks pretty cool. If that's how you normally interact on this project, I'm up.

scripts/build.sh
```diff
@@ -19,11 +19,11 @@ browserify \
   $(find src/test -name '*.js')
 
 # Create dist/pileup.js
+#  -g uglifyify \
```
armish: any reason why you are getting rid of minification?

src/main/data/SamRead.js
```diff
@@ -61,16 +63,39 @@ class SamRead /* implements Alignment */ {
    */
   constructor(buffer: ArrayBuffer, offset: VirtualOffset, ref: string) {
     this.buffer = buffer;
-    this.offset = offset;
+
```
armish: @selkovjr: are these changes in SamRead.js an artifact of a problematic merge? If so, would you mind doing a rebase on the master branch? If not, I am not sure if I understand the reasoning behind these changes.

armish:
thanks again for these changes, @selkovjr! Looking at them in more details now, I see that many of these changes are probably due to a faulty merge which reverted some of the files back to their older versions. 

CGI-based solution is indeed a great practical one, but I don't think we should include it in the distribution as pileup.js is meant to be a pure JS library. Having said that, we should definitely support setting a cap to the number of reads to be loaded by default and discourage against it when possible. Your samtools-based solution is a nice workaround the problem and I think we should mention this in the documentation if you would like to set up a stable repo for it (otherwise, happy to link to your fork from the documentation). 

@ihodes, what do you think?

armish:
We already have some code visualizing the number of requests made and we should be able to re-utilize this for showing a warning when that number exceed the user-defined one. 

Context: https://github.com/hammerlab/pileup.js/commit/ffa26eaaa4448c0ce5c2fcba05a943e37a9c8eea

ihodes:
I'm going to close out this issue for now; we can reopen as necessary.

selkovjr:
Sure. Sorry I have been so slow. It took me a couple months just to get
started, then a few more to figure out where I wanted to go with it, and
finally I was hit with the strictest of deadlines: I don't work there
anymore.

The result was quite useful, though, and it has been tested by several
people from whom I collected feedback. Now I am negotiating code release
with the company I worked for. Some of the code didn't make it to the
branch I forked. Meanwhile, I can summarize it from memory. A lot of it was
specific to Ion Torrent, but from a functional standpoint, some of the
adaptations around Torrent data can be generally useful. So I will list
both accomplishments and requests for new functionality that I had on my
to-do list.

1. CGI access was a big hit, and think I was able to do it right, although
I also think the solution was made possible and even easy by the way people
in the R&D part of the company shared data. Everybody's home directory was
mounted off the same root, with universal read access. So I could write a
simple CGI script to reach anybody's BAMs anywhere on the system by
absolute path. The ugly state-of-the art was that people either had to
download the 20-50GB BAMs to their local stations or do arcane ssh mounts
to access them. That's done, and all relevant code is in the branch.

2. We need to be ready for huge coverage. Some pileups I worked with last
reached 50,000x in well-covered regions. The original pileup widget would
render a blank canvas when fed too many reads. I tracked that down to the
maximum render area in pixels that is browser-specific (although the most
important browsers have the same maximum height of 32767 pixels). I solved
that by adding a hard size limit, which is not great, but better than a
blank canvas. The fix is in the branch. The proper fix should involve some
sort of paging. Another obvious fix is downsampling, but that is a regular
CGI parameter in the access script, anyway.

3. A peek into the future reveals that we'll be increasingly working with
complex ordering and grouping in the pileups; for example, grouping reads
by molecular tag. For a small number of tags, just ordering with a custom
sort function and inserting dividers or brackets might do well; maybe a
number of scroll-synchronized pileups will work (one per family). For
larger quantities (a thousand families, with possibly a thousand reads or
more in some families) -- I don't know. For now, the solution was to give
them a filter option in the CGI accessor, so they can create a web page
with several widgets, each holding one family. This approach will soon hit
a limit in the number of canvas elements on a page, but at least it's
workable in the short term. It's just an idea -- no code to show here.
Maybe it's something to mind in relation to paging: page by quantity as
well as by attribute, possibly punctuating with page breaks.

4. Detailed view. In the master branch, you have an alert with raw read
data that opens when you click on a read. I had a lot more data to show, so
I experimented with a proper (modal) pop-up, but quickly realized its
shortcomings, the worst of which was that it obstructed the pileup view. In
the end, I settled on a split-page resizable view similar to Web Inspector.
You click on a read, and the page splits, with the bottom part showing the
details of the read. Hit Escape or push the close button, and it goes away.
The content of the detailed view is assembled from the data about the read
already available through the pileup source, supplemented by more data from
another CGI request. The centerpiece of the detailed view is the flowgram
showing the raw signal, phase-corrected sequence prediction, and residuals.
No sensible investigation of errors was possible without a flowgram. Until
I put things together in a single web widget, people had been analyzing
flowgrams either by looking at tables with numbers or plotting them in R,
while sticking a finger into an IGV pileup -- a terribly complicated
routine I myself have not been able to master. I have yet no code to show
for that (pending release) but there is a hook in my branch to render a
detailed view via CGI.

5. Now that I have a pop-up or a split, I want to know which read I clicked
on to open. My branch has the code to highlight a read on a click.

6. Related to the above, but not implemented, people wanted a balloon to
show on hover with basic information about the alignment. For example the
last question asked was, "Can I see all insertions spelled out?"

7. Also related, implemented, but not perfectly, is the ability to position
the pileup to a selected read. I have a new method in
PileupTrack, scrollTo(read index), which allows me to bring the selected
read into view. There are two reasons for that. One is that I might want to
open a new pileup with a focus on that read (which is a common need when
documenting cases). Another one arises from the detailed view obscuring the
reads in the lower half of the pileup view when the page is split between
the two views. If the read I click on becomes invisible, it must be somehow
strolled into visibility. I currently scroll it to the 3rd position from
top, but that does not always work well. At any rate, that logic is part of
the Torrent-specific code I don't yet have, but I think the ability to home
in on a read must must be expressed by the pileup API. So it is in my
branch.

8. The center-column marker in the main branch is somewhat useful as a
navigation aid, but we need more. I have added a permanent marker to
highlight a position of interest that is supplied as a CGI argument to the
wrapper script (the one that is used to invoke the pileup and put it on the
page -- either stand-alone or embedded). That marker remains attached as
the pileup is scrolled sideways.

9. Cursors. Now that I have a detailed view with a couple linear elements
in it -- the flowgram and the expanded alignment -- I want cursors in all
those elements to be synchronized on hover. The co-ordinate mapping is
application-specific, to be sure, but the pileup widget needs at least
enough code to render a cursor and to move it around. I am not sure I've
don it right: I ended up re-rendering the whole thing to update the cursor.
It seems to work reasonably well; if there is a smarter way to do it, I
would like to know. The cursor now in my branch is a single-letter box
highlighting a position in the selected read. The critique I got for that
was that the cursor must span the entire pileup vertically, including the
reference track.

This is a rough-and-ready catalogue of what I have done or intended to do.
I will update you if I recall anything else, and I will push for the
release of the Torrent-specific, so it can at least be used as a working
model.

Issue #406, opened 2016-03-20
Not able to load BAM using pileupjs


deepaksibar:
I am trying to load my BAM file using pileupjs. But BAM is not getting loaded and following error thrown by pileupjs library.

q.js:161 Uncaught Monster request: Won't fetch Infinity bytes from ./data/bam/B907_2__reHeader_Sorted.bam

I am trying with demo code given on hammerlab website and using my BAM file (B907_2__reHeader_Sorted.bam and its index file) instead of demo BAM file.

Please let me know why i am getting this error and how it can be fixed ??

armish:
Duplicate of #405. 

@deepaksibar: if you don't mind, let's continue the conversation under #405.

PR #407, opened 2016-04-04
Better handle BAI chunk sizes (fixes #405 and #406)

### LINKED ISSUE #405: Unable to load my local bam file in pileupjs
I have installed WAMP stack and I am trying to load a BAM file but I am getting below error.

_"q.js:161 Uncaught Monster request: Won't fetch Infinity bytes from ./test-data/molBarcode.bam"_

Am I doing something wrong here?

below is the code snippet from data.js
 
var bamSource = pileup.formats.bam({
  url: './test-data/molBarcode.bam',
  indexUrl: './test-data/molBarcode.bam.bai'
   });

armish:
For reasons unknown to me, some BAI files seem to have contigs of size 0. This was forcing pileup.js to try to fetch `Infinity` bytes and eventually fail. This change captures the edge case and caps the chunk size as a fix to issue #405 and #406.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/407)

### DIFF
--- a/src/main/data/bai.js
+++ b/src/main/data/bai.js
@@ -22,6 +22,7 @@ import VirtualOffset from './VirtualOffset';
 // winds up saving time to do a fast pass over the data to compute them. This
 // allows us to parse a single contig at a time using jBinary.
 function computeIndexChunks(buffer) {
+  var BLOCK_SIZE = 65536;
   var view = new jDataView(buffer, 0, buffer.byteLength, true /* little endian */);
 
   var minBlockIndex = Infinity;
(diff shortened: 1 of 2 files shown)


armish:
looks like the `weigh_in` task doesn't work anymore. So I am removing that step from the CI while at it.

src/main/data/bai.js
```diff
@@ -40,9 +40,7 @@ function computeIndexChunks(buffer) {
     if (n_intv) {
       var offset = VirtualOffset.fromBlob(view.getBytes(8), 0),
           coffset = offset.coffset + (offset.uoffset ? 65536 : 0);
-      if (coffset) {
-        minBlockIndex = Math.min(coffset, minBlockIndex);
-      }
+      minBlockIndex = coffset ? Math.min(coffset, minBlockIndex) : 65536;
```
ihodes: What's the significance of 65536? (Maybe make a constant here instead of this magic number?)

ihodes: Also, wherever this 0 was turning into infinity, may be good to assert/comment that it can't be 0 at that point.

armish: Done. Thanks @ihodes! Merging this one in.

Issue #410, opened 2016-05-06
pileup.js (difficulties to load susScr3.2bit)


marcguellc:
Hey,
I'm experiencing difficulties to load our .2bit file ([susScr3](http://hgdownload.cse.ucsc.edu/goldenPath/susScr3/bigZips/susScr3.2bit)). See my data.js below. I have no problem in loading any of the test .2bit files( test.2bit, http://www.biodalliance.org/datasets/hg19.2bit). I cannot understand what is the problem. I'd would really appreciate any advice. This is the error I get:

![image 6](https://cloud.githubusercontent.com/assets/3690357/15075102/9f521e48-1370-11e6-8734-bd26928a1b18.png)

Thanks a lot!

my data.js file

```
# Some data for the demo.

var sources = [
  {
    viz: pileup.viz.genome(),
    isReference: true,
    data: pileup.formats.twoBit({
      url: 'http://hgdownload.cse.ucsc.edu/goldenPath/susScr3/bigZips/susScr3.2bit'
    }),
    name: 'Reference'
  },
  {
    viz: pileup.viz.scale(),
    name: 'Scale'
  },
  {
    viz: pileup.viz.location(),
    name: 'Location'
  }
];

var range = {contig: 'chr1', start: 1, stop: 1000};
```

armish:
Thanks a lot for reporting this problem, Marc! Since our 2bit parser was built from bottom up, we sometimes have edge cases that we have never thought of. Sorry that you hit one such case.

I will take a closer look at how to fix this in a few days and will make a new release. Will also update this thread as soon as I am done. Stay tuned.

armish:
@marcguellc: OK, this should be fixed with the newest version of the pileup.js: [v0.6.5](https://www.npmjs.com/package/pileup). You can update your working copy via `npm update pileup`.

Thanks again for reporting this issue and please let me know if have trouble getting this to work.

PR #411, opened 2016-05-08
Better handle 2bit files with huge headers

armish:
In the context of #410, it turned out that some 2bit files can have headers that are too long to be captured  by a single request of 16 KB (our earlier heuristic) and this can cause parsing issues on our hand. The same goes with the contig information (4 KB was failing to capture the complete contig metadata).

This adds support for continuously asking for bigger chunks if the initial request returns an incomplete header. The size requests are multiplied every time a parsing error occurs but these are also capped to 1 MB to prevent unnecessary network traffic.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/411)


src/main/data/TwoBit.js
```diff
@@ -124,6 +136,32 @@ function markUnknownDNA(basePairs: Array<string>, dnaStartIndex: number, sequenc
...
+ * If the intial request returns an incomplete header and hence the
+ * parsing fails, we next try doubling the requested size.
+ * The size of the request is capped with `untilSize` so that
+ * we don't start asking for MBs of data for no use.
+ * Instead we we throw an error if we reach the cap,
+ * potentially meaning a corrupted 2bit file.
+*/
+function retryRemoteGet(remoteFile: RemoteFile, start: number, size: number, untilSize: number, promisedFunc: Function) {
+  return remoteFile.getBytes(start, size).then(promisedFunc).catch(error => {
+    if(error.constructor == RangeError) {
```
ihodes: What is throwing the range error?

src/main/data/TwoBit.js
```diff
@@ -124,6 +136,32 @@ function markUnknownDNA(basePairs: Array<string>, dnaStartIndex: number, sequenc
...
+      // Do not attempt to download more than `untilSize`
+      if(size > untilSize) {
+        throw `Couldn't parse the header ` +
+              `from the first ${size} bytes of the file. ` +
+              `Corrupted 2bit file?`;
+      }
+      return retryRemoteGet(remoteFile, start, size*2, untilSize, promisedFunc);
```
ihodes: Could you, instead of re-requesting from the `start`, request from `start + size` to `size * 2`, saving some bandwidth and hopefully speeding things up?

ihodes:
Thanks for handling this! Couple little questions.

src/main/data/TwoBit.js
armish: Those errors get thrown by JBinary's `read` method within our `parse{Header, SequenceRecord}`. When the inferred record size is bigger than the actual buffer passed to that `read` function, you get a `RangeError`.

src/main/data/TwoBit.js
armish: That is definitely possible, but that solution would require passing around partial buffers and concatenating them, which is not that hard to implement; but, I thought the way it is right now is  easier to read/interpret and cleaner. Normally, I don't expect this logic to be called that often, which also made me think that wasting tens or hundreds of KBs for rare edge cases shouldn't be that big of a problem.

So yeah — I chose readability over performance, but if you think that might eventually bite us in a way I don't foresee, I am happy to optimize those network requests.

ihodes: Definitely--I'm fine with leaving it.

src/main/data/TwoBit.js
ihodes: Could be nice to catch that error within the parser and throw a more descriptive error; would that be reasonable for this PR? The `RangeError` in the context of this function is pretty confusing (I was wondering why we'd catch a `RangeError` and then try to get a longer chunk).

armish: That is a good idea :+1:

armish:
having trouble resolving an Heisenbug that shows up in the Travis environment. Apologies up front for the upcoming force pushes as a way to investigate this. Will update the thread as soon as I fix this.

ihodes:
Nice, LGTM!

Issue #415, opened 2016-06-09
destroy method doesn't clean everything


piotr-gawron:
I've noticed that calling destroy method doesn't clean everything. When pileup object is created then destroyed and finally resize event is triggered, you can see on the console an error: "Uncaught Error: Minified exception occurred; use the non-minified dev environment for the full error message and additional helpful warnings.". 

Here is a jsfiddle showing the issue:
http://jsfiddle.net/udq10buk/3/

And summary code that causes errors:

```javascript
var p = pileup.create(element, {
  range: range,
  tracks: sources
});
p.destroy();
window.dispatchEvent(new Event('resize'));
```

As far as I could tell the problem is that destroy method doesn't remove onresize event handler from window object.
The errror doesn't crash application. However, I think it should be fixed ;).

PR #418, opened 2016-06-10
Fix on cleaning during destroy method

piotr-gawron:
This small pull request fixes issue that I reported here: https://github.com/hammerlab/pileup.js/issues/415

During pileup destroy call, listener added by VisualizationWrapper at initialization stage is removed.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/418)

### DIFF
--- a/src/main/VisualizationWrapper.js
+++ b/src/main/VisualizationWrapper.js
@@ -33,6 +33,7 @@ class VisualizationWrapper extends React.Component {
   props: Props;
   state: {width: number; height: number};
   hasDragBeenInitialized: boolean;
+	onResizeListener;
 
   constructor(props: Object) {
     super(props);
@@ -52,7 +53,10 @@ class VisualizationWrapper extends React.Component {
   }
 
   componentDidMount(): any {
-    window.addEventListener('resize', () => this.updateSize());
+		//local copy of the listener, so we can remove it
+		//when pileup is destroyed
+		this.onResizeListener = () => this.updateSize();
+    window.addEventListener('resize', this.onResizeListener);
     this.updateSize();
 
     if (this.props.range && !this.hasDragBeenInitialized) this.addDragInterface();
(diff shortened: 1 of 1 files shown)


armish:
@piotr-gawron: this looks great! Thanks for the fix. 

My only comment would be regarding your use of tabs for the indentas. If you can replace those with spaces, this should be good to go :+1:

piotr-gawron:
@armish: indent is fixed.

I will probably add some more pull request this month. I hope you will find them usefull as well :).

I'm not sure if this is the right place to put some of my remarks about the code. If you want to move it somewhere else or delete permanently just do it ;). Anyway here it is (please remember that I'm not a JS developr - I'm Java dev and I work with JS when I have to ;)):
* writing the code is a bit difficult due to long build time - when I was debugging test cases to speed up development I was modifing built code, but this of course is tricky because it's translated JS5 and everything is compiled into one huge file and you need to move your code later on to original source files. If you know about faster way to do it, please put it somewhere.
* some of your test suits have code explicitely in the body, I think that this should be moved to before method, so it's run only when tests from the suit is executed. I discovered it when I was running single test. I found out that some strange parts of the code were also executed
* cleaning after test suite is executed should be improved. Right now when test case fails you will not call pileup.destroy which might lead to unsuccessful exec of other test not related to the original code. In general it should be done either in some finally statement (I'm not sure if there is something like this in mocha) or in "after" method that will be called after test was executed

armish:
@piotr-gawron: these are all good points. I tried to recapitulate your last comments within #421. Feel free to add to that issue if you have further comments.

As for the slowness of the build: I suggest you use our `npm run watch` workflow instead of going down the full `build` route every time you make a change. The watch script only transpiles the files that have been changed and the build is relatively faster — hence, it allows you to, for example, change a file, wait for a few seconds and refresh the test page. In the past, we talked about testing out different build options; but, we never were able to get to that. 

BTW: please keep these PRs coming! We really appreciate your contribution. And, since you mentioned there might be more of these, I will wait until you are done with this batch of changes and then make a release.

Issue #416, opened 2016-06-09
Visualization of reference genome issue (GenomeTrack issue)


piotr-gawron:
I found out that when pileup is created with larg initial range when you zoom in the reference genome track doesn't change into colorfull ACGT letters/rectangles. It stays blank.

Here is a jsfiddle with the code that shows the issue:
http://jsfiddle.net/Lf16furq/1/

The summary code that shows the problem:

```javascript
//initial large range
var initialRange = {contig: 'chr1', start: 104788594, stop: 124271154};
var element = document.getElementById('pileup');
var p = pileup.create(element, {
  range: initialRange,
  tracks: sources
});
//now we zoom in (this is done programatically, 
//but it's the same when user zooms in using '+' button),
//so we should see single nucleotides, however nothing changes
p.setRange({contig: 'chr1', start: 114529834, stop: 114529884});
```

The workaround that I found is rather silly: always initialize pileup.js with small range. And right after initialization change it to the range that you wanted:

```javascript
//initial large range
var initialRange = {contig: 'chr1', start: 104788594, stop: 124271154};

///some artifitial short range that we will use in initialization step
var shortArtifitialRange = {contig: 'chr1', start: 114529834, stop: 114529884};
var element = document.getElementById('pileup');
var p = pileup.create(element, {
  range: shortArtifitialRange,
  tracks: sources
});
//and now switch to original initial range
p.setRange(initialRange);

//and zoom-in on reference genome track works at it's supposed to
p.setRange({contig: 'chr1', start: 114529834, stop: 114529884});
```

PR #419, opened 2016-06-13
Devel genome track bug

piotr-gawron:
This pull request is a patch that fixes bug reported here: https://github.com/hammerlab/pileup.js/issues/416

Here is brief description of the bug source:
When TwoBitDataSource object was requested to obtain data from huge range it checked if the region was already fetched ('covered') from remote file. If the chunk wasn't fetched from remote file it marked the region as fetched and then tried to fetch it. However, in the fetch method there was additional check that compared the requested range size with maximum allowed fetch range size. When request was too big it silently skipped fetching.
I moved "marking regions as covered" after range size checks.

I provided two test cases for the bug:
* low level unit test for TwoBitDataSource which shows the flow in patched file
* test case that shows how the bug affected the whole pileup vizualization component (with some gui checks as well)

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/419)

### DIFF
--- a/src/main/sources/TwoBitDataSource.js
+++ b/src/main/sources/TwoBitDataSource.js
@@ -68,8 +68,14 @@ var createFromTwoBitFile = function(remoteSource: TwoBit): TwoBitSource {
   function fetch(range: ContigInterval) {
     var span = range.length();
     if (span > MAX_BASE_PAIRS_TO_FETCH) {
+			//informa that we won't fetch the data
+      o.trigger('newdatarefused', range);
       return Q.when();  // empty promise
     }
+		//now we cam add region to covered regions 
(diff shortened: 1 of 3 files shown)


armish:
thanks, @piotr-gawron! This also looks great and thanks for taking the time to add those tests in. Once the indents are fixed, I will merge this in and make a new release (w/ #418). 

It would be great if you can squash these commits into a single one with a short description followed by your long one, but if it is too much trouble; I can do that on my own as well.

piotr-gawron:
@armish  I changed tabs into spaces. 

Regarding single commit, I think you can use my first commit desc as merge description.

armish:
Looks great! Merging this in :+1:

PR #420, opened 2016-06-14
Dynamically generated vcf content

piotr-gawron:
This pull request contains simple functionality that allows to visualize vcf files that were dynamically created in memory by javascript code.

Ideally it would be great to have possibility to create dynamic content for variant track. Here I provide a simple tool that allows it to some extend. When creating variant track there is possibility to intialize it with the content of vcf file instead of url. 

Of course it should be used for small dynamic content.

Some technical comments:
* I implemented LocalStringFile that has the same interface as RemoteFile. This class is initialized with content of the file in a String
* There is "abstract" class that is a super class for LocalStringFile and RemoteFile. However, due to some lint and facebook/flow restrictions implementation of this class is not perfect (it allows instantiation, method signature doesn't contain return type)
* There are tests for new functionalities
* Implementation of LocalStringFile will probably break when input string contain characters that need more than one byte, but for VCF files it's enough

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/420)

### DIFF
--- a/src/main/AbstractFile.js
+++ b/src/main/AbstractFile.js
@@ -1,0 +1,44 @@
+/**
+ * AbstractFile is an abstract representation of a file. There are two implementation:
+ * 1. RemoteFile  - representation of a file on a remote server which can be
+ * fetched in chunks, e.g. using a Range request.
+ * 2. LoclaStringFile is a representation of a file that was created from input string. 
+ * Used for testing and small input files.
+ * @flow
+ */
+'use strict';
+
+import Q from 'q';
+
+
+class AbstractFile {
+  constructor() {
+    //how to prevent instantation of this class???
+    //this code doesn't pass npm run flow
+//    if (new.target === AbstractFile) { 
+//      throw new TypeError("Cannot construct AbstractFile instances directly");
+//    }
+  }
+
+  getBytes(start: number, length: number): Q.Promise<ArrayBuffer> {
+    throw new TypeError("Method getBytes is not implemented");
+  }
+
+  // Read the entire file -- not recommended for large files!
+  getAll(): Q.Promise<ArrayBuffer> {
+    throw new TypeError("Method getAll is not implemented");
+  }
+
+  // Reads the entire file as a string (not an ArrayBuffer).
+  // This does not use the cache.
+  getAllString(): Q.Promise<string> {
+    throw new TypeError("Method getAllString is not implemented");
+  }
+
+  // Returns a promise for the number of bytes in the remote file.
+  getSize(): Q.Promise<number> {
+    throw new TypeError("Method getSize is not implemented");
+  }
+}
+
+module.exports = AbstractFile;
--- a/src/main/AbstractFile.js
+++ b/src/main/AbstractFile.js
@@ -2,7 +2,7 @@
  * AbstractFile is an abstract representation of a file. There are two implementation:
  * 1. RemoteFile  - representation of a file on a remote server which can be
  * fetched in chunks, e.g. using a Range request.
- * 2. LoclaStringFile is a representation of a file that was created from input string. 
+ * 2. LocalStringFile is a representation of a file that was created from input string. 
  * Used for testing and small input files.
  * @flow
  */
--- a/src/main/AbstractFile.js
+++ b/src/main/AbstractFile.js
@@ -8,7 +8,7 @@
  */
 'use strict';
 
-import Q from 'q';
+//import Q from 'q';
 
 class AbstractFile {
   constructor() {
@@ -19,23 +19,23 @@ class AbstractFile {
 //    }
   }
 
-  getBytes(start: number, length: number): Q.Promise<ArrayBuffer> {
+  getBytes(start: number, length: number):Object {//: Q.Promise<ArrayBuffer> {
     throw new TypeError("Method getBytes is not implemented");
   }
 
   // Read the entire file -- not recommended for large files!
-  getAll(): Q.Promise<ArrayBuffer> {
+  getAll():Object {//: Q.Promise<ArrayBuffer> {
     throw new TypeError("Method getAll is not implemented");
   }
 
   // Reads the entire file as a string (not an ArrayBuffer).
   // This does not use the cache.
-  getAllString(): Q.Promise<string> {
+  getAllString():Object {//: Q.Promise<string> {
     throw new TypeError("Method getAllString is not implemented");
   }
 
   // Returns a promise for the number of bytes in the remote file.
-  getSize(): Q.Promise<number> {
+  getSize():Object {//: Q.Promise<number> {
     throw new TypeError("Method getSize is not implemented");
   }
 }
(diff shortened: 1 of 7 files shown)


src/main/AbstractFile.js
```diff
@@ -0,0 +1,43 @@
+/**
+ * AbstractFile is an abstract representation of a file. There are two implementation:
+ * 1. RemoteFile  - representation of a file on a remote server which can be
+ * fetched in chunks, e.g. using a Range request.
+ * 2. LoclaStringFile is a representation of a file that was created from input string.
```
armish: Locla -> Local

src/main/sources/VcfDataSource.js
```diff
@@ -91,13 +92,16 @@ function createFromVcfFile(remoteSource: VcfFile): VcfDataSource {
   return o;
 }
 
-function create(data: {url:string}): VcfDataSource {
+function create(data: Object): VcfDataSource {
```
armish: How about using optional field annotations instead? Maybe something like this:

```javascript
function create(data: {url?: string, content?: string}): VcfDataSource {
  var {url, content} = data;
  if (url) {
    return createFromVcfFile(new VcfFile(new RemoteFile(url)));
  } else if (content) {
    return createFromVcfFile(new VcfFile(new LocalStringFile(content)));
  }
  // If no URL or content is passed, fail
  throw new Error(`Missing URL or content from track: ${JSON.stringify(data)}`);
}
```

src/test/data/vcf-test.js
```diff
@@ -51,3 +52,58 @@ describe('VCF', function() {
     });
   });
 });
+describe('VCF from string content', function() {
```
armish: great that you have this test to check the local file implementation, but I have bad feelings about embedding the contents of the `snv.vcf` here, as this will come with a necessity of sync'ing that file and this test content later on. How about abstracting things a bit more and extracting the content of that file via `RemoteFile` instead:

```javascript
/* @flow */
'use strict';

import {expect} from 'chai';

import VcfFile from '../../main/data/vcf';
import ContigInterval from '../../main/ContigInterval';
import RemoteFile from '../../main/RemoteFile';
import LocalStringFile from '../../main/LocalStringFile';

describe('VCF', function() {
  describe('should respond to queries', function() {
    var testQueries = (vcf) => {
      var range = new ContigInterval('20', 63799, 69094);
      return vcf.getFeaturesInRange(range).then(features => {
        expect(features).to.have.length(6);

        var v0 = features[0],
            v5 = features[5];

        expect(v0.contig).to.equal('20');
        expect(v0.position).to.equal(63799);
        expect(v0.ref).to.equal('C');
        expect(v0.alt).to.equal('T');

        expect(v5.contig).to.equal('20');
        expect(v5.position).to.equal(69094);
        expect(v5.ref).to.equal('G');
        expect(v5.alt).to.equal('A');
      });
    };

    var remoteFile = new RemoteFile('/test-data/snv.vcf');

    it('remote file', function() {
      var vcf = new VcfFile(remoteFile);
      testQueries(vcf);
    });

    it('local file from string', function() {
      return remoteFile.getAllString().then(content => {
        var localFile = new LocalStringFile(content);
        var vcf = new VcfFile(localFile);
        testQueries(vcf);
      });
    });
  });

  it('should add chr', function() {
    var vcf = new VcfFile(new RemoteFile('/test-data/snv.vcf'));
    var range = new ContigInterval('chr20', 63799, 69094);
    return vcf.getFeaturesInRange(range).then(features => {
      expect(features).to.have.length(6);
      expect(features[0].contig).to.equal('20');  // not chr20
      expect(features[5].contig).to.equal('20');
    });
  });

  it('should handle unsorted VCFs', function() {
    var vcf = new VcfFile(new RemoteFile('/test-data/sort-bug.vcf'));
    var chr1 = new ContigInterval('chr1', 1, 1234567890),  // all of chr1
        chr5 = new ContigInterval('chr5', 1, 1234567890);
    return vcf.getFeaturesInRange(chr1).then(features => {
      expect(features).to.have.length(5);
      return vcf.getFeaturesInRange(chr5);
    }).then(features => {
      expect(features).to.have.length(5);
    });
  });
});
```

I think this is simpler and easier to maintain. Let me know what you think.

armish:
@piotr-gawron: again — this is great, thanks! I have a few comments and suggestions here and there before we merge this in. Let me know if you need more clarification on any of these.

src/main/sources/VcfDataSource.js
piotr-gawron: sure,

As I mentioned. I'm not JavaScript developer and Ecmascript 2015 is definitely new to me. I didn't know that I can do that.

src/test/data/vcf-test.js
piotr-gawron: Indeed, it's better.

For my defence I will only say that this was the first thing that I wrote in your project and I didn't know a lot about the code that you have, and didn't think later on about refactoring it...

piotr-gawron:
@armish I agree with all of them. Thanks for checking my code. I inlcuded your comments in the PR (I'm not sure if you wanted me to do it or not :)).

armish:
@piotr-gawron: cool - thanks for addressing those comments!

Two minor things to address before we call this a merge:
 - with the recent addition, looks like we ended up with two `create` functions within `VcfDataSource.js`. Can you remove the old one with the `Object` signature and leave only the recent addition?
 - the new `create` uses tabs for indents; would be great if you can convert them to spaces while at it.

src/main/sources/VcfDataSource.js
armish: No worries - I, also, had to check the `flow-type` reference and try a few things before I get that right. Not the most obvious solution ;)

src/main/sources/VcfDataSource.js
```diff
@@ -91,13 +92,27 @@ function createFromVcfFile(remoteSource: VcfFile): VcfDataSource {
...
+    return createFromVcfFile(new VcfFile(new RemoteFile(url)));
   }
+  if (content!==null && content!== undefined) {
+    return createFromVcfFile(new VcfFile(new LocalStringFile(content)));
+  }
+  throw new Error(`Missing URL from track: ${JSON.stringify(data)}`);
+}
 
-  return createFromVcfFile(new VcfFile(new RemoteFile(url)));
+function create(data: {url?: string, content?: string}): VcfDataSource {
```
armish: thanks for adding this implementation in. Can you remove the old one from the source code?

src/main/sources/VcfDataSource.js
```diff
@@ -91,13 +92,27 @@ function createFromVcfFile(remoteSource: VcfFile): VcfDataSource {
...
+  var {url, content} = data;
+	if (url) {
```
armish: tabs -> spaces

piotr-gawron:
@armish I hope I fixed it properly this time.

armish:
@piotr-gawron: 👍

denisemauldin:
@piotr-gawron How is this functionality used?  Can you add example code?  I have a JSON file that I'd like to load into pileup.js and it seems like the easiest way is to transform the JSON into a VCF File and then load it into pileup.js in memory.  Thanks!

Issue #417, opened 2016-06-09
GeneTrack zooming issue


piotr-gawron:
I found an issue when working with GeneTrack.
When pileup object is created genes are visualized properly on track: pileup.viz.genes.
However, when user click zoom out button ("-") list of visible genes is not updated (genes that were not visible, but should become visible after zoom out don't appear).

Here is a simple jsfiddle example that simulates clicks on zoom out button:

http://jsfiddle.net/bnLqdwg8/2/

PR #422, opened 2016-06-15
Fix of the gene-track zoom issue

piotr-gawron:
Here is a patch that fixes https://github.com/hammerlab/pileup.js/issues/417 issue.

Short desription of the problem:
When passing information between GeneTrack and BigBedDataSource at some point intervals were mixed up (interval that was requested, and interval for which data was obtained). This resulted in skipping data that didn't match to both of the intervals.

@armish I mentioned more pull requests. This is the last for now. I plan to do at least one more (add onclick events on variants), but I don't have time now. Hopefully I will get back to it in late July.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/422)


armish:
@piotr-gawron: Looks great — thanks for the fix!

armish:
@piotr-gawron: FYI -> new release: https://github.com/hammerlab/pileup.js/releases/tag/v0.6.6

thanks a lot for all these improvements/fixes!

Issue #414, opened 2016-06-07
Way to force rendering of pileup component


piotr-gawron:
I'm writing quite complex application. I was planning to use pileup.js for vizualization gene variants data. However, I encountered some rendering issues when creating pileup object before attaching it to DOM.

For instance, the code below will not render properly:

``` javascript
var element = document.getElementById('pileup');

var newElement = document.createElement("div");

var p = pileup.create(newElement, {
  range: range,
  tracks: sources
});

element.appendChild(newElement);
```

The reason (as far as I understand) is that when pileup object is created the newElement doesn't belong to DOM, so it's not intialized/rendered. This make sense - it's a performance issue. However, when I add newElement to the DOM I expect the object to be initialized and rendered properly, but it doesn't happen. Is there a method that I can call to force rendering? Or can this issue be fixed in any other way?

The only workaround that I found is to trigger resize event on window object:

``` javascript
window.dispatchEvent(new Event('resize'));
```

but this might not be a cross-brower approach (I checked it only on firefox 46.0.1 and chrome 50.0.2661.102)

Here is a jsfiddle that visualizes the problem: http://jsfiddle.net/40ygmtd8/5/

danvk:
It probably is rendered, it's just that unstyled `div`s have zero height. There's no way that I'm aware of for a JS library to learn about its `div` being resized (the Google Maps API has the same issue, for example). So dispatching a resize event isn't a bad workaround. Or perhaps pileup could expose a `resize()` method of its own.

piotr-gawron:
Hm... I thought the same and I did some tests: I created a div with fixed size, but I got the same result:

tak a look here: 
http://jsfiddle.net/40ygmtd8/6/ (the div width and height are: 500px)

When you investigate elments in chrome or firefox you will see that canvas elements (used for visualization) have non zero width, height, but there is no content and the size is different.

Anyway, I found this rendering issue also in other situations, so it would be nice to have a method that would force pileup to re-render everything.

piotr-gawron:
I think the source of this issue is that in tracks (GeneTrack.js, GenomeTrack.js, etc.) the component where content is visualized is retrieved using: [ReactDOM.findDOMNode](https://facebook.github.io/react/docs/top-level-api.html#reactdom.finddomnode) method. This will not work if a div where we want to visualize our data is not in a DOM.

PR #425, opened 2016-06-29
Issue 414

piotr-gawron:
This PR fixes issue #414. However, there are problems with tests now. Tests works perfectly fine when
they are run in a browser, but when test are run in a console there are errors. The errors can be summarized as: object 'document' has no function 'contains'.

Can you help me with tests?

Summary of my PR: when pileup is created with 'element' that doesn't belong to DOM, observer object is created. This observer observes document object. When 'element' is added to document (or to any (grand)child) of document pileup tracks are forced to rerender by changing updateSize state of ReactElements.

Sorry for the commits comments but I have problems with lint and git in general ;/

---

This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/425)

### DIFF
--- a/src/main/pileup.js
+++ b/src/main/pileup.js
@@ -80,6 +80,37 @@ function create(elOrId: string|Element, params: PileupParams): Pileup {
       ReactDOM.render(<Root referenceSource={referenceTrack.source}
                             tracks={vizTracks}
                             initialRange={params.range} />, el);
+
+  //if the element doesn't belong to document DOM observe DOM to detect
+  //when it's attached
+  var observer = null;
+  if (!document.contains(el)) {
+    observer = new MutationObserver(function(mutations) {
+      mutations.forEach(function(mutation) {
+        if (mutation.type === 'childList') {
+          var added= false;
+          for (var i=0; i<mutation.addedNodes.length;i++) {
+            if (mutation.addedNodes[i]===el) {
+              added = true;
+            }
+          }
+          if (added) {
+            if (reactElement) {
+              reactElement.setState({updateSize:true});
+            } else {
+              throw 'ReactElement was not initialized properly';
+            }
+          }
+        }
+      });
+    });
+    // configuration of the observer:
+    var config = { attributes: true, childList: true, characterData: true, subtree: true };
+
+    // start observing document
+    observer.observe(document, config)
+  }
+
   return {
     setRange(range: GenomeRange) {
       if (reactElement === null) {
@@ -104,6 +135,11 @@ function create(elOrId: string|Element, params: PileupParams): Pileup {
       reactElement = null;
       referenceTrack = null;
       vizTracks = null;
+
+      // disconnect observer if it was created
+      if (observer!=null) {
+        observer.disconnect();
+      }
     }
   };
 }
--- a/src/main/pileup.js
+++ b/src/main/pileup.js
@@ -108,7 +108,7 @@ function create(elOrId: string|Element, params: PileupParams): Pileup {
     var config = { attributes: true, childList: true, characterData: true, subtree: true };
 
     // start observing document
-    observer.observe(document, config)
+    observer.observe(document, config);
   }
 
   return {
@@ -137,7 +137,7 @@ function create(elOrId: string|Element, params: PileupParams): Pileup {
       vizTracks = null;
 
       // disconnect observer if it was created
-      if (observer!=null) {
+      if (observer !== null && observer !== undefined) {
         observer.disconnect();
       }
     }
--- a/src/main/pileup.js
+++ b/src/main/pileup.js
@@ -91,7 +91,10 @@ function create(elOrId: string|Element, params: PileupParams): Pileup {
         if (mutation.type === 'childList') {
           var added= false;
           for (var i=0; i<mutation.addedNodes.length;i++) {
-            if (mutation.addedNodes[i]===el) {
+            //when added element is element where we visualize pileup
+            //or it contains element where we visualize pileup
+            //then we will have to update component
+            if (mutation.addedNodes[i]===el || mutation.addedNodes[i].contains(el)) {
               added = true;
             }
           }
@@ -106,7 +109,7 @@ function create(elOrId: string|Element, params: PileupParams): Pileup {
       });
     });
     // configuration of the observer:
-    var config = { childList: true, subtree: true };
+    var config = {attributes: true, childList: true, characterData: true, subtree: true};
 
     // start observing document
     observer.observe(document, config);
--- a/src/main/pileup.js
+++ b/src/main/pileup.js
@@ -84,7 +84,8 @@ function create(elOrId: string|Element, params: PileupParams): Pileup {
   //if the element doesn't belong to document DOM observe DOM to detect
   //when it's attached
   var observer = null;
-  if (!document.contains(el)) {
+
+  if (!document.body.contains(el)) {
     observer = new MutationObserver(function(mutations) {
       mutations.forEach(function(mutation) {
         if (mutation.type === 'childList') {
@@ -105,7 +106,7 @@ function create(elOrId: string|Element, params: PileupParams): Pileup {
       });
     });
     // configuration of the observer:
-    var config = { attributes: true, childList: true, characterData: true, subtree: true };
+    var config = { childList: true, subtree: true };
 
     // start observing document
     observer.observe(document, config);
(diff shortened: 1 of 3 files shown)


src/main/pileup.js
```diff
@@ -80,6 +80,37 @@ function create(elOrId: string|Element, params: PileupParams): Pileup {
       ReactDOM.render(<Root referenceSource={referenceTrack.source}
                             tracks={vizTracks}
                             initialRange={params.range} />, el);
+
+  //if the element doesn't belong to document DOM observe DOM to detect
+  //when it's attached
+  var observer = null;
+  if (!document.contains(el)) {
```
armish: TIL that `document.contains` is not a standard thing:  http://stackoverflow.com/questions/20557115/mocha-casperjs-headless-testing-document-contains-undefined

but `if (!document.body.contains(el)) {` works fine as suggested in that SO answer. This should fix the issue with the headless tests.

src/main/pileup.js
```diff
@@ -80,6 +80,37 @@ function create(elOrId: string|Element, params: PileupParams): Pileup {
...
+              reactElement.setState({updateSize:true});
+            } else {
+              throw 'ReactElement was not initialized properly';
+            }
+          }
+        }
+      });
+    });
+    // configuration of the observer:
+    var config = { attributes: true, childList: true, characterData: true, subtree: true };
```
armish: if we are only looking for mutations of type `childList`, do we need all the other ones? I guess `subtree` is relevant, but if not let's drop the others and only keep the `childList`.

src/main/pileup.js
```diff
@@ -80,6 +80,37 @@ function create(elOrId: string|Element, params: PileupParams): Pileup {
...
+          for (var i=0; i<mutation.addedNodes.length;i++) {
+            if (mutation.addedNodes[i]===el) {
+              added = true;
+            }
+          }
+          if (added) {
+            if (reactElement) {
+              reactElement.setState({updateSize:true});
+            } else {
+              throw 'ReactElement was not initialized properly';
```
armish: can you give me an example why we might need this error? Looks like this type of an error is out of the scope for pileup.js?

src/main/pileup.js
```diff
@@ -80,6 +80,37 @@ function create(elOrId: string|Element, params: PileupParams): Pileup {
...
+            }
+          }
+        }
+      });
+    });
+    // configuration of the observer:
+    var config = { attributes: true, childList: true, characterData: true, subtree: true };
+
+    // start observing document
+    observer.observe(document, config);
```
armish: I have no experience with the `MutationObserver` API, but do you think observing the whole document DOM tree will result in some performance issues for us? Do you have an idea about how we can only watch some relevant elements and not the whole document?

src/main/pileup.js
```diff
@@ -80,6 +80,37 @@ function create(elOrId: string|Element, params: PileupParams): Pileup {
       ReactDOM.render(<Root referenceSource={referenceTrack.source}
                             tracks={vizTracks}
                             initialRange={params.range} />, el);
+
+  //if the element doesn't belong to document DOM observe DOM to detect
```
armish: How about using some underscore utilities to make this code more concise? For example:

```javascript
  //if the element doesn't belong to document DOM observe DOM to detect
  //when it's attached
  var observer = null;
  if (!document.body.contains(el) && reactElement) {
    observer = new MutationObserver(mutations => {
      var updateSize = !_.chain(mutations)
        .filter(m => m.type === 'childList' && _.contains(m.addedNodes, el))
        .isEmpty()
        .value();
      reactElement.setState({updateSize});
    });

    // start observing document
    observer.observe(document, {childList: true});
  }

/* ... */

      // disconnect observer if it was created
      if (observer) {
        observer.disconnect();
      }
```

armish:
thanks @piotr-gawron! See my inline comments/questions regarding some functionality and performance issues.

It would also be great to be able to replicate this issue within a unit test, but I do understand it might be tricky/unfeasible to do so. But having such as a case might help us do some benchmarking and performance monitoring.

On a side note: do you think it would be better if we expose an `updateSize` method within the main pileup object and document this edge case in case somebody else hits a similar issue? That way, we don't have to think about how this might affect other users in a broader sense.

src/main/pileup.js
piotr-gawron: I dissagree with this answer at SO. Here you have better source:
https://developer.mozilla.org/en-US/docs/Web/API/Node/contains

It is a standard, the only remark is that IE doesn't support it for document but only for document.body.

I can change it to document.body if it solves the problem.

Anyway, I took it from this answer at SO:
http://stackoverflow.com/questions/5629684/how-to-check-if-element-exists-in-the-visible-dom/5629730#5629730

src/main/pileup.js
piotr-gawron: Well, I don't like when the code fails to do what it's supposed to do and don't give any hint what went wrong. 
It helps debugging, testing etc.

src/main/pileup.js
piotr-gawron: I also don't have any experience. I found an article here:
https://hacks.mozilla.org/2012/05/dom-mutationobserver-reacting-to-dom-changes-without-killing-browser-performance/
They mention it as the proper way to do it when performance is important (I know it's 4 years old, but couldn't find better way to do that).

I was thinking about watching proper elements, but I couldn't find any better solution. My first idea was quite opposit - watch our node element if it's parent has changed, but I couldn't make it work (even though that MutationObserver in theoury allows to observe modification of node attributes - maybe parent is not treated as an attribute...).
This is why I added this observer only when it's not in DOM. Maybe we should disconnect observer when element is attached to DOM.

src/main/pileup.js
piotr-gawron: Probably you are right, I will check it.

src/main/pileup.js
piotr-gawron: You can do that, but for me this code is really hard to read and understand. This later on will influence time spent on debuging and improving code.

But this is my opinion. Maybe for Javascript developers this syntax is obvious and easy to read :).

src/main/pileup.js
armish: as an alternative, you can do:

```javascript
var contains = document.contains || document.body.contains;
if (!contains(el)) {
...
}
```

PhantomJS framework can sometimes surface issues like this; but, the pros still outweigh its cons, so we are still sticking to it. Also, any such issue will affect people trying to render views in headless mode, so it would be great if we address this even though it is a weird workaround.

src/main/pileup.js
armish: thanks a lot!

src/main/pileup.js
armish: I say we keep it if you hit this problem when trying to solve this issue. The check for the reactElement happens deep within that `if` block, though. Maybe it makes sense to move it to the top?

src/main/pileup.js
armish: Fair enough. If you don't have any performance issues in your complex application after this change, it is unlikely that we will have that for other simpler applications. Thanks for the pointer!

src/main/pileup.js
armish: your version is definitely more explicit, but let's delegate some of the logic to underscore for clarity. The for-loop-based check is a pretty common pattern, so I think it is safe to replace that with the `filter` method. What we can do is to add a few sentences in between the chain so that it becomes more obvious for future developers taking a look at the code.

src/main/pileup.js
piotr-gawron: this doesn't work: I get some strange Invalid invocation error in Chrome, Firefox reports different error...

But I fixed it by replacing: `document.contains` with `document.body.contains`

src/main/pileup.js
piotr-gawron: We can move it up. But it must be inside MutationObserver constructor.
There might be some issues due to lifecycle of react component (I'm not familiar with it at all): 
* consider object creation
* object destruction - which method will be called when - maybe observer still will be running when react element is gone??

Anyway, where do you want to move it? Inside forEach loop? (line 90)

src/main/pileup.js
piotr-gawron: I will leave it to you. I don't know underscore notation.

piotr-gawron:
@armish
Sorry for long answer - it's vacations time and I wasn't available for the last 2 weeks.

piotr-gawron:
@armish I created this PR almost a year ago. Do you plan to merge it? Or there are some issues with my code?

armish:
hey @piotr-gawron - sorry that this fell through the cracks and I wasn't able to get back to you with my second round of reviews. I originally wanted to do some style/type cleaning up on your code before merging in, but for the last few months, I barely have the additional time to channel into pileup maintenance.

Let me just take a quick look at your PRs and I will merge them in if there are no apparent red flags. 

Apologies for the delay.

PR #426, opened 2016-07-04
Disabled scrolling to negative regions

akmorrow13:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/426)

### DIFF
--- a/src/main/Root.js
+++ b/src/main/Root.js
@@ -50,6 +50,10 @@ class Root extends React.Component {
   }
 
   handleRangeChange(newRange: GenomeRange) {
+    // Do not propagate negative ranges
+    if (newRange.start < 0) {
+      newRange.start = 0;
+    }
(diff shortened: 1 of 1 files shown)


src/main/Root.js
```diff
@@ -50,6 +50,10 @@ class Root extends React.Component {
   }
 
   handleRangeChange(newRange: GenomeRange) {
+    // Do not propigate negative ranges
+    if (newRange.start < 0) {
```
armish: thanks for this change, @akmorrow13!

Do you think this would better behave in terms of the UI if we do `newRange.start = 0` instead of just doing nothing (i.e. `return`)?

akmorrow13:
Good call @armish . I have made those updates to now set the start to 0 if it is negative.

armish:
@akmorrow13: cool - thanks for the fix! Will make a minor release with this fix in a minute :+1:

PR #429, opened 2016-07-12
Added support for remote requests

akmorrow13:
* Modified test scripts for faster test compilation
* Added RemoteRequest and tests to support generic http requests

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/429)

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,8 @@
     "react": "^0.14.0",
     "react-dom": "^0.14.0",
     "shallow-equals": "0.0.0",
-    "underscore": "^1.7.0"
+    "underscore": "^1.7.0",
+    "memory-cache": "0.1.6"
   },
   "devDependencies": {
     "arraybuffer-slice": "^0.1.2",
--- a/src/main/RemoteRequest.js
+++ b/src/main/RemoteRequest.js
@@ -1,0 +1,88 @@
+/**
+ * RemoteRequest is a generic endpoint for serving http requests. RemoteRequest
+ * handles json data, which is specified by genomic range (contig, start, stop)
+ *
+ * @flow
+ */
+'use strict';
+
+import Q from 'q';
+import ContigInterval from './ContigInterval';
+
+class RemoteRequest {
+  url: string;
+  cache: Object;
+  numNetworkRequests: number;  // track this for debugging/testing
+
+  constructor(url: string) {
+    this.cache = require('memory-cache');
+    this.url = url;
+    this.numNetworkRequests = 0;
+  }
+
+  get(contig: string, start: number, stop: number): Q.Promise<Object> {
+    var length = stop - start;
+    if (length <= 0) {
+      return Q.reject(`Requested <0 interval (${length}) from ${this.url}`);
+    }
+
+    // First check the cache.
+    var contigInterval = new ContigInterval(contig, start, stop);
+    var buf = this.cache.get(contigInterval);
+    if (buf) {
+      return Q.when(buf);
+    }
+
+    // Need to fetch from the network.
+    return this.getFromNetwork(contig, start, stop);
+  }
+
+  /**
+   * Request must be of form "url/contig?start=start&end=stop"
+  */
+  getFromNetwork(contig: string, start: number, stop: number): Q.Promise<Object> {
+    var length = stop - start;
+    if (length > 5000000) {
+      throw `Monster request: Won't fetch ${length} sized ranges from ${this.url}`;
+    }
+    var xhr = new XMLHttpRequest();
+    var endpoint = this.getEndpointFromContig(contig, start, stop);
+    xhr.open('GET', endpoint);
+    xhr.responseType = 'json';
+    xhr.setRequestHeader('Content-Type', 'application/json');
+
+    return this.promiseXHR(xhr).then(json => {
+      // extract response from promise
+      var buffer = json[0];
+      var contigInterval = new ContigInterval(contig, start, stop);
+      this.cache.put(contigInterval, buffer);
+      return buffer;
+    });
+  }
+
+  getEndpointFromContig(contig: string, start: number, stop: number): string {
+    return `${this.url}/${contig}?start=${start}&end=${stop}`;
+  }
+
+  // Wrapper to convert XHRs to Promises.
+  // The promised values are the response (e.g. an ArrayBuffer) and the Event.
+  promiseXHR(xhr: XMLHttpRequest): Q.Promise<[any, Event]> {
+    var url = this.url;
+    var deferred = Q.defer();
+    xhr.addEventListener('load', function(e) {
+      if (this.status >= 400) {
+        deferred.reject(`Request for ${url} failed with status: ${this.status} ${this.statusText}`);
+      } else {
+        deferred.resolve([this.response, e]);
+      }
+    });
+    xhr.addEventListener('error', function(e) {
+      deferred.reject(`Request for ${url} failed with status: ${this.status} ${this.statusText}`);
+    });
+    this.numNetworkRequests++;
+    xhr.send();
+    return deferred.promise;
+  }
+}
+
+module.exports = RemoteRequest;
--- a/src/test/RemoteRequest-test.js
+++ b/src/test/RemoteRequest-test.js
@@ -1,0 +1,67 @@
+/** @flow */
+'use strict';
+
+import {expect} from 'chai';
+
+import sinon from 'sinon';
+
+import RemoteRequest from '../main/RemoteRequest';
+import RemoteFile from '../main/RemoteFile';
+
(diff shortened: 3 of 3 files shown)


src/main/RemoteRequest.js
```diff
@@ -0,0 +1,107 @@
...
+ * @flow
+ */
+'use strict';
+
+import Q from 'q';
+
+type Chunk = {
+  start: number;
+  stop: number;
+  buffer: Object; // TODO: more efficient data store
```
akmorrow13: Is there a more efficient way we can store data generically?

src/main/RemoteRequest.js
```diff
@@ -0,0 +1,107 @@
...
+      var buffer = json[0];
+
+      var newChunk = { start, stop, buffer};
+      this.chunks.push(newChunk);
+      return buffer;
+    });
+  }
+
+  getEndpointFromContig(contig: string, start: number, stop: number): string {
+    return this.url + "/" + contig + "?start=" + start + "&end=" + stop + "&key=" + this.key;
```
akmorrow13: This is a slightly messy way to get the endpoint. Is it acceptable to hardcode a key on the endpoint? This would be used for file identification.

src/main/RemoteRequest.js
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  // Wrapper to convert XHRs to Promises.
+  // The promised values are the response (e.g. an ArrayBuffer) and the Event.
+  promiseXHR(xhr: XMLHttpRequest): Q.Promise<[any, Event]> {
```
akmorrow13: This function is almost a replicate from RemoteFile. Maybe we could make some linker class for common functionality between the files. I am not sure the best way to do this.

src/main/RemoteRequest.js
armish: @akmorrow13: you mentioned this yesterday, but I am a little bit confused about why we need that `key` field at all. I understand the G4GH API doesn't describe how to handle the file identifier, but can't we force the server side to handle requests like:

`http://domain.com/bam/file/{key}/?start={start}&end={end}`

instead of 

`http://domain.com/bam/file/?start={start}&end={end}&key={key}`

I think we favored the former one in [Cycledash](https://github.com/hammerlab/cycledash) — right, @ihodes?

package.json
```diff
@@ -31,7 +31,8 @@
     "flow": "flow status",
     "flow-check": "flow check",
     "coverage": "./scripts/generate-coverage.sh",
-    "build": "./scripts/build.sh"
+    "build": "./scripts/build.sh",
+    "quick-build": "./scripts/quick-build.sh"
```
armish: @akmorrow13: just curious about why you need this. `npm run watch` should be fast enough for quick turnarounds while developing, but if that's not useful maybe we should split the `build.sh` as you did.

src/main/RemoteRequest.js
armish: Not off the top of my head :(

cc: @ihodes @danvk

danvk: I think I'm missing some context here. How is this distinct from `src/main/RemoteFile.js`?

akmorrow13: The main difference is that RemoteFile is file oriented, and much of its functionality is focused around parsing and storing bytes. RemoteRequest deals with json, not byte data. Do you think these differences can be reconciled as one class?

src/main/RemoteRequest.js
akmorrow13: I was being silly. http://domain.com/bam/file/{key}/?start={start}&end={end} should work just fine. The user can be in charge of providing the descriptor in the URL for the datasource.

package.json
akmorrow13: Ah wonderful! I didn't know the purpose of watch.sh. I will remove quick-build.sh.

src/main/RemoteRequest.js
akmorrow13: We could also use a module for a cache (https://github.com/ptarjan/node-cache). However, I am not sure if they use any smart compression to decrease client side blowup. How do people feel about this?

danvk: pileup.js has always drawn a distinction between data classes and source classes:
https://github.com/hammerlab/pileup.js/blob/master/DESIGN.md

The data classes (e.g. `bam.js`) are responsible for issuing network requests and parsing bytes into structures. The source classes (e.g. `BamDataSource.js`) are responsible for caching features in a range (a `ContigInterval`) and exposing a synchronous API.

This class might be more like a source class than a data class. I think [`GA4GHDataSource`][1] is a close analogue to what you're doing here.

[1]: https://github.com/hammerlab/pileup.js/blob/55dc2213a9ccf6bdb336f402a42e0632776f8e7c/src/main/sources/GA4GHDataSource.js

akmorrow13: Exactly. We talked on Monday with you folks about replacing GA4GHDataSource with a generic access point. If you think it would be better in the folder sources I can move it there. But generally I was thinking we leave RemoteRequest in the main folder like RemoteFile and make specific datasources (like GA4GHDataSource, CoverageDataSource, etc.) in the sources folder which use RemoteRequest.js.  See https://github.com/akmorrow13/pileup.js/tree/master/src/main/sources <https://github.com/akmorrow13/pileup.js/tree/master/src/main/sources>.

As far as caching goes, RemoteFile has a generic cache as well that keeps Data sources from accessing the network. GA4GHDatatSource does not have a cache so we could swap RemoteRequest in here for the cache.

Best,

Alyssa Morrow
akmorrow@berkeley.edu <mailto:akmorrow@berkeley.edu>
414-254-6645

armish: I think the folder structure you have in mind makes sense @akmorrow13  if you will be using this class as a utility class rather than extending it. If the latter, it might make sense to make it more abstract and move it under the `source` folder.

src/main/RemoteRequest.js
armish: I am in favor of extracting this method as a utility class or creating an abstract class with this method in it where both `Remote*` classes extend it. Depending on your decision about where to put this class and how you will use it, feel free to implement it in a way that is the easiest for you.

src/main/RemoteRequest.js
```diff
@@ -0,0 +1,87 @@
...
+      // extract response from promise
+      var buffer = json[0];
+      var contigInterval = new ContigInterval(contig, start, stop);
+      this.cache.put(contigInterval, buffer);
+      return buffer;
+    });
+  }
+
+  getEndpointFromContig(contig: string, start: number, stop: number): string {
+    return this.url + "/" + contig + "?start=" + start + "&end=" + stop;
```
armish: how about making this a bit more consice with [ES6's template feature](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)? e.g.:

```javascript
   return `${this.url}/${contig}?start=${start}&end=${stop}`;
```

armish:
a few minor comments, but otherwise this LGTM, @akmorrow13! You can address the issues about abstraction and the location of this new file later on if you are planning to implement all those `*DataSource` classes within a different PR.

Let me know if you want me to merge this or wait until you are done with the other classes as well.

src/main/RemoteRequest.js
```diff
@@ -0,0 +1,87 @@
+/**
+ * RemoteRequest is used to for http requests on a remote server which can be
+ * fetched in chunks, e.g. using a Range request.
```
danvk: Is this copy/paste from [`RemoteFile`][1]? I don't think it describes this class.

[1]: https://github.com/hammerlab/pileup.js/blob/55dc2213a9ccf6bdb336f402a42e0632776f8e7c/src/main/RemoteFile.js

src/main/RemoteRequest.js
```diff
@@ -0,0 +1,87 @@
...
+    return this.getFromNetwork(contig, start, stop);
+  }
+
+  /**
+   * Request must be of form "url/contig?start=start&end=stop"
+  */
+  getFromNetwork(contig: string, start: number, stop: number): Q.Promise<Object> {
+    var length = stop - start;
+    if (length > 50000000) {
+      throw `Monster request: Won't fetch ${length} bytes from ${this.url}`;
```
danvk: bytes or base pairs?

akmorrow13:
Ok I fixed some small comments and am now good with the merge! We can wait to abstract it. I will be working on coverage next but this way @ryan-williams can take a look if he is interested.

armish:
@akmorrow13: sorry for the delay in getting back to this. Let's take care of the coverage improvements in a different branch — it would make the review process easier for all of us.

Merging this in!

PR #430, opened 2016-07-12
Working draft: BigWig support

ryan-williams:
This doesn't work yet! Just going to use this PR to collaborate with @armish while we work on this today.

This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/430)


ryan-williams:
I'm working on rebasing out that merge conflict atm

ryan-williams:
OK, it's more up to date, I did some extra Bed+Wig refactoring to share some common code paths, have the Bed support working again, but Wig support is still not complete / compilable.

ryan-williams:
One weird thing is `npm run lint` is giving me a couple of apparent false positives:

```
src/main/viz/CoverageTiledCanvas.js: line 14, col 8, 'DepthCache' is defined but never used.

src/main/viz/WiggleTrack.js: line 14, col 8, 'BigWigSource' is defined but never used.
```

Those "unused" classes are declared as part of a `type` and as an instance var of a `class`, though not used directly anywhere yet.

ryan-williams:
Also flow is very unhappy, I'll look at that tmrw.

armish:
hmm... those lint issues are all new to me. Maybe it is time we made the switch to eslint. I will experiment with that and update you ⌛

ryan-williams:
OK it now recursively fetches R-tree indices for the whole file and also for each zoom level.

I've added a BigWig-test that references `bigWigExample.bw` which I've refrained from checking in thus far because it is ~80MB; if ppl don't mind such a test file being added lmk. It is also served from http://genome.ucsc.edu/goldenPath/help/examples/bigWigExample.bw, though swapping that in as the URL in the test doesn't work because genome.ucsc.edu doesn't set the appropriate CORS headers.

Anyway, next I'll add a data-block cache and then some code to compute stats for arbitrary buckets based on the underlying zoom-levels, then that will be ready to feed to a view.

src/main/viz/WiggleTrack.js
```diff
@@ -0,0 +1,182 @@
...
+
+import _ from 'underscore';
+import React from 'react';
+import d3utils from './d3utils';
+import dataCanvas from 'data-canvas';
+import shallowEquals from 'shallow-equals';
+import style from '../style';
+
+import WiggleCache from './WiggleCache';
+import BigWigSource from '../sources/BigWigDataSource';
```
armish: `import type BigWigSource from '../sources/BigWigDataSource';` and the lint error should be gone

src/main/viz/CoverageTiledCanvas.js
```diff
@@ -0,0 +1,163 @@
...
+
+import _ from 'underscore';
+
+import type {DataCanvasRenderingContext2D} from 'data-canvas';
+import ContigInterval from '../ContigInterval';
+import TiledCanvas from './TiledCanvas';
+import scale from '../scale';
+import type {BinSummary} from './CoverageCache';
+import style from '../style';
+import DepthCache from './DepthCache';
```
armish: `import type DepthCache from './DepthCache'; ` (see my other comment)

armish:
@ryan-williams: I am sitting on the fence about that 80MB test file. Do you think we can somewhat filter that file to reduce its size? Would be great it to limit the file size to <2MB, but I understand this might be challenging.

ryan-williams:
Yea I agree 80MB is too big; I thought about it and I should be able to easily make test .bw's that test all the cases but are much smaller.

The example has a R-tree branching-factor of 256 and a lot of data per "block" so to get any non-leaf nodes in the tree you need a large amount of data, but those parameters can be tweaked in `wigToBigWig` so I'll go that route!

superbobry:
You might want to have a look at our implementation of [BigWig](https://github.com/JetBrains-Research/big/blob/master/src/main/kotlin/org/jetbrains/bio/big/BigWig.kt) and [BigBed](https://github.com/JetBrains-Research/big/blob/master/src/main/kotlin/org/jetbrains/bio/big/BigBed.kt) in addition to the UCSC codebase. I've found UCSC code a bit hard to comprehend due to general C verbosity and the codestyle.

PR #433, opened 2016-07-21
Pileup coverage

akmorrow13:
- Added coverage
- Extended coverage to sample by region size

---

This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/433)

### DIFF
--- a/src/main/RemoteRequest.js
+++ b/src/main/RemoteRequest.js
@@ -28,31 +28,30 @@ class RemoteRequest {
     return new ContigInterval(range.contig, newStart, newStop);
   }
 
-  getFeaturesInRange(range: ContigInterval<string>): Q.Promise<Object> {
+  getFeaturesInRange(range: ContigInterval<string>, modifier: string = ""): Q.Promise<Object> {
     var expandedRange = this.expandRange(range);
-    return this.get(expandedRange.contig, expandedRange.start(), expandedRange.stop());
+    return this.get(expandedRange, modifier);
   }
 
-  get(contig: string, start: number, stop: number): Q.Promise<Object> {
-    var length = stop - start;
+  get(range: ContigInterval<string>, modifier: string = ""): Q.Promise<Object> {
+
+    var length = range.stop() - range.start();
     if (length <= 0) {
       return Q.reject(`Requested <0 interval (${length}) from ${this.url}`);
+    } else if (length > 5000000) {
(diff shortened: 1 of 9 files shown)


src/main/RemoteRequest.js
```diff
@@ -11,57 +11,62 @@ import ContigInterval from './ContigInterval';
...
-      var contigInterval = new ContigInterval(contig, start, stop);
-      this.cache.put(contigInterval, buffer);
-      return buffer;
+      return json[0];
     });
   }
 
-  getEndpointFromContig(contig: string, start: number, stop: number): string {
-    return `${this.url}/${contig}?start=${start}&end=${stop}`;
+  getEndpointFromContig(contig: string, start: number, stop: number, modifier: string = ""): string {
```
akmorrow13: The modifier is used for coverage bytes per bin. This could also be extend to include filters for variants, etc.

src/main/sources/CoverageDataSource.js
```diff
@@ -0,0 +1,103 @@
...
+
+    // Check if this interval is already in the cache.
+    if (cache.coversRange(interval)) {
+      return Q.when();
+
+    }
+
+    // modify endpoint to calculate coverage using binning
+    var basePairsPerBin = ResolutionCache.getResolution(interval.interval);
+    var endpointModifier = `binning=${basePairsPerBin}`;
```
akmorrow13: @david4096 is trying to develop a standardized GA4GH schema for coverage. Does anyone have opinions on how this would look? Here, we have 
'/coverage/contig?start=start&end=end&binning=bin' where binning is the number of bytes per bin.

src/main/ResolutionCache.js
```diff
@@ -0,0 +1,168 @@
...
+ *
+ * @flow
+ */
+'use strict';
+
+import _ from 'underscore';
+import Interval from './Interval';
+import ContigInterval from './ContigInterval';
+
+class ResolutionCache<T: Object> {
```
akmorrow13: I am not a huge fan of the terminology 'resolution' or 'binning' to refer to requests that explicitly specify resolution. @ryan-williams would this be useful for bringing in multiple resolution coverage in #430?

akmorrow13:
The next step here would be to add or modify the existing CoverageTrack to work with the new remote endpoint. However, I am not sure of the best way to do this.

Issue #434, opened 2016-09-12
Load alignments text jumpiness


Georgehe4:
The pileup page jumps down when new alignments are loaded because of the [Loading Alignments] text box.

PR #435, opened 2016-09-13
Resolve load alignment text jump issues

Georgehe4:
Address #434

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/435)

### DIFF
--- a/style/pileup.css
+++ b/style/pileup.css
@@ -196,6 +196,8 @@
   display: flex;
   align-items: center;
   justify-content: center;
+  position: relative;
+  top: 30px;
 }
 .pileup .network-status-message {
   padding: 4px 8px;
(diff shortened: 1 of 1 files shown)


armish:
thanks a lot for this fix, @Georgehe4! 

Just curious: why not add those styles into the [pileup.css](https://github.com/hammerlab/pileup.js/blob/master/style/pileup.css), instead? It would be great to continue decoupling styles from the JSX as much as possible.

Georgehe4:
@armish Updated! Yep I agree with the decoupling.

armish:
@Georgehe4: awesome — thank you! Merging this in and will be cutting a new release soon 🙇

PR #441, opened 2016-11-02
Visualization of genome track

piotr-gawron:
This PR contains 3 main things:
1) fix on the issue with popups when clicking on variant in Variant track - right now when you click on variant in variant track you will get an error message in a console (supposed outcome should be popup with information about gene variant)
2) popups for gene variants use [bootstrap](https://react-bootstrap.github.io/)  [model window](https://react-bootstrap.github.io/components.html#modals)  to present detailed information about variant from vcf file
3) when visualizing vcf file I added possibility to use variant bar height as an indicator of allel frequency - this is something requested by biologist at our center - this is optional, not used by default

Few comments regaring changes:
* I decided to use [bootstrap](https://react-bootstrap.github.io/) as a popup implementation because it was the easiest chocie for me. If you don't like the solution then I can use any other if you provide me with links to some examples/documentation. One of the reasons why you might not like it is the fact that it requires bootstrap css to be included (if not the modal popup is useless). Of course this can be included in pileup css if you want to keep the whole css stuff in one file, but I'm not sure what are you rstandards here.

* user (programmer) can decide if he wants to use allel frequency as a hight of variant bar by providng options parameter when creating gene variant view. This requires a lot of passing config parameters: pileup->Root->VisualizationWrapper->VariantTrack. I couldn't find better solution, so if you know one please let me know and I will refactor.
```
{
    viz: pileup.viz.variants(),
    data: pileup.formats.vcf({
      url: '/test-data/snv.chr17.vcf'
    }),
    options: {variantHeightByFrequency: true},
    name: 'Variants'
  }
```

* I'm not sure why some of my commits from June are in this PR (I couldn't find any changes that wasn't introduced this week in the code)

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/441)


src/main/data/vcf.js
```diff
@@ -41,12 +43,28 @@ function extractLocusLine(vcfLine: string): LocusLine {
 
 function extractVariant(vcfLine: string): Variant {
   var parts = vcfLine.split('\t');
+  var frequency = null;
+  if (parts.length>=7){
+    var params = parts[7].split(';');
+    for (var i=0;i<params.length;i++) {
+      var param = params[i];
+      if (param.startsWith("AF=")) {
```
armish: I think `AF` is an almost standard annotation for a majority of the VCF variants out there, but do you know if there are other abbreviations that people use as a synonym? If so, it might worth looking for those.

src/main/data/vcf.js
```diff
@@ -41,12 +43,28 @@ function extractLocusLine(vcfLine: string): LocusLine {
 
 function extractVariant(vcfLine: string): Variant {
   var parts = vcfLine.split('\t');
+  var frequency = null;
+  if (parts.length>=7){
```
armish: What if this is a multi-sample VCF and the first sample is not relevant for the user? I am asking this; because I know that mutect can sometimes order the samples in a way that  the normal sample comes before the tumor one.

piotr-gawron:
I understand - when library is light is much easier to maintain (and it's more flexible).

I will reimplement it accordingly - will remove bootstrap from dependencies and correct example with the registered callback function.

However, there is still a question of passing options to pileup. Do you agree with the methodology that I used? (passing configuration options from top classes to bottom classes).

src/main/data/vcf.js
piotr-gawron: As far as I know vcf can contain multiple variants regarding single nucleotide. It can be achieved in two ways:
1) one entry in vcf can correspond to more than one variant, take a look at this entry:
```20 1234567 microsat1 GTC G,GTCT 50 PASS NS=3;DP=9;AA=G GT:GQ:DP 0/1:35:4 0/2:17:2 1/1:40:3```
2) two entries can refer to the same region in the genome (I think this is what you mentioned in the comment)

Regarding the first issue - there is nothing you can do in term of visualization apart from providing information in the popup. We could think about coloring or some fancy way of highlighting the situation, but in my opinion it's not intuitive.

When you have two (or more) overlaying variants I would also suggest to put it in the popup. I will fix my PR to handle this situation (right now you have only first element that match click in the popup).

If the behaviour is not the one user expected I think the best way to go is suggest user to filter data from vcf file and provide filtered results.

Anyway, this reminds me about one other issue: when you present gene variant it should span through the whole modified region. Right now every variant is visualized by rectangle on the first nucleotide only.

File: src/main/data/vcf.js [hunk 3]
```diff
@@ -41,12 +43,28 @@ function extractLocusLine(vcfLine: string): LocusLine {
...
+    }
+  }
 
   return {
     contig: parts[0],
     position: Number(parts[1]),
+    id: parts[2],
     ref: parts[3],
     alt: parts[4],
+    significantFrequency: frequency,
```
piotr-gawron: The reason is that single vcf entry can contain information about more than one gene variant:
``
20 1110696 rs6040355 A G,T 67 PASS NS=2;DP=10;AF=0.333,0.667;AA=T;DB
``
In such situation we need to decide which allel frequency value use for the visualization.
I know about two strategies, but there might be more: 
* take the smalles
* take the biggest (this one is currently implemented)

The alternative to `significantFrequency` field is to store all the 'allel frequency' values and allow user to decide. But again we have to agree about the possible strategies.

armish: That makes sense. Can we name those two alternatives as minor vs major allele frequency just to be in line with the common population biology terminology? It would also be great to mention  why pick either of these in the comments right at the top of the code block so that future developers won't get confused.

armish:
thanks for the clarifications, @piotr-gawron! I think the way the way you implemented things for passing options to visualization tracks look reasonable. If this proves not that useful in the future, we can change the things to better accommodate other use cases.

src/main/data/vcf.js
piotr-gawron: I wouldn't go that far. In most cases users know what data they visualize and when they know that there are two overlapping data sets they should separate them in different tracks (at least this is what I would do).

piotr-gawron:
Let me first explain why travis failed:
`flow` library doesn't support get/set properties in ES2015, you can find more information about it here: https://github.com/facebook/flow/issues/572

I used get method because I needed to specify default prop value for allel frequency strategy and this is the right way to do. Alternatively you can do something like: http://stackoverflow.com/a/36778923/1127920. However this extracts specificiation of the prop default outside of the class and I don't like the approach.

I noticed also one more issue with your code: 
you shouldn't specify prop value for typechecking. Here you can find article about propTypes and defaultProps standard used in react: https://facebook.github.io/react/docs/typechecking-with-proptypes.html
The way you did it prevented setting default value to the prop.

piotr-gawron:
@armish I see that my PR has some pending requestes changes. Can you tell me what is it, or point to the point on github, because I cannot find it...

examples/index.html
```diff
@@ -3,6 +3,7 @@
 <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
 <link rel="stylesheet" href="../style/pileup.css" />
 <link rel="stylesheet" href="demo.css" />
+<link rel="stylesheet" href="https://maxcdn.bootstrapcdn.com/bootstrap/latest/css/bootstrap.min.css">
```
armish: can you remove this stylesheet since we are not making use of it anymore?

armish: Never mind - let me do that to save some time.

piotr-gawron:
@armish 
Thanks for merging. Do you plan by any chance to release it?
It would make my life easier if I could include npm package with merged changes :).

PR #442, opened 2016-11-04
Fixed collapsed zoom button bug in Electron.

chgibb:
Under Electron "+" and "-" do not render on the zoom in and out buttons. This addresses this. Not tested for breakage in browser.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/442)

### DIFF
--- a/src/main/Controls.js
+++ b/src/main/Controls.js
@@ -112,8 +112,8 @@ class Controls extends React.Component {
         <input ref='position' type='text' />{' '}
         <button className='btn-submit' onClick={this.handleFormSubmit.bind(this)}>Go</button>{' '}
         <div className='zoom-controls'>
-          <button className='btn-zoom-out' onClick={this.zoomOut.bind(this)}></button>{' '}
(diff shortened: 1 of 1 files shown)


chgibb:
@armish I'll give it a look on Monday. Thanks for the quick reply!

chgibb:
@armish other than this warning, there's nothing in the console.
```
'CanvasRenderingContext2D.webkitImageSmoothingEnabled' is deprecated. Please use 'CanvasRenderingContext2D.imageSmoothingEnabled' instead.
```

Do you know what the default search path is for pileup.css? I'm requiring the dist/pileup.js file generated by npm run build through browserify.

chgibb:
@armish poking through your pileup.electron repo I figured out the path. Thanks for the feedback!

armish:
@chgibb: glad that old hackathon project proved useful :P

By the way, really curious to learn more about your tool at some point when you think that is ready to share. It is always fun to see pileup.js in action.

chgibb:
@armish looking forward to sharing as it progresses :)

PR #443, opened 2016-11-05
Fix for zoomed in letter visibility

snaheth:
This is done by adjusting the y-value parameter in the fillText call to be height/2 and adjusting a font size.

Previously, a font size of 24px was used once multiple zooming-in (more zooming in for pictures below) was done. This was problematic since the canvas has a height of only 20px. I thought we could go about resolving this by either resizing the canvas when a font size of 24px was wanted, or we could simply lower the font size. I took the latter approach.

Make sure to do `npm run build` before running `npm run http-server`

## Results

**Few zooms in:**
screen shot 2016-11-05 at 12 41 18 am

**More zooming in:**
screen shot 2016-11-05 at 12 41 09 am

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/443)

### DIFF
--- a/src/main/viz/GenomeTrack.js
+++ b/src/main/viz/GenomeTrack.js
@@ -53,7 +53,7 @@ function renderGenome(ctx: DataCanvasRenderingContext2D,
         // We only push objects in the text case as it involves creating a
         // new object & can become a performance issue.
         // 0.5 = centered
-        ctx.fillText(letter, scale(1 + 0.5 + pos), height - 1);
+        ctx.fillText(letter, scale(1 + 0.5 + pos), height / 2);
       } else {
         if (pxPerLetter >= style.COVERAGE_MIN_BAR_WIDTH_FOR_GAP) {
           // We want a white space between blocks at this size, so we can see
--- a/src/main/style.js
+++ b/src/main/style.js
@@ -19,7 +19,7 @@ module.exports = {
   },
 
   // Styles for base pairs which are rendered as letters
-  LOOSE_TEXT_STYLE: `24px 'Helvetica Neue', Helvetica, Arial, sans-serif`,
+  LOOSE_TEXT_STYLE: `14px 'Helvetica Neue', Helvetica, Arial, sans-serif`,
   TIGHT_TEXT_STYLE: `bold 12px 'Helvetica Neue', Helvetica, Arial, sans-serif`,
 
   // Gene track


src/main/viz/GenomeTrack.js
```diff
@@ -53,7 +53,7 @@ function renderGenome(ctx: DataCanvasRenderingContext2D,
         // We only push objects in the text case as it involves creating a
         // new object & can become a performance issue.
         // 0.5 = centered
-        ctx.fillText(letter, scale(1 + 0.5 + pos), height - 1);
+        ctx.fillText(letter, scale(1 + 0.5 + pos), height / 2);
```
snaheth: The argument changed was the y-coordinate. The previous y coordinate was height - 1, which was placing the text at the lower end of the canvas. By setting it to height / 2, the text was centered vertically.

akmorrow13: Ah ok.

akmorrow13:
Looks good. Make sure to squash the 2 commits.  @armish , does this change look ok?

src/main/viz/GenomeTrack.js
snaheth: Commits squashed!

armish: Thanks for the clean up!

armish: also I don't think passing the center would cause the text to be vertically centered. The alignment of the text should depend on the combination of a few things: http://www.w3schools.com/tags/canvas_textalign.asp

Or am I missing something here?

snaheth: You're right. Hmm. I'm not even sure then. Let me look at this more.

snaheth: I'm on a Mac OS 10.11.2 using Chrome/54.0.2840.87. Regarding your screen capture, I saw the top cutoff as well on specific letters (G and T mostly), which is why using 13px may be the way to go - if we can't get the text to center vertically.

akmorrow13: @snaheth , if you change the screen width it will cause text to render differently. I would recommend making the browser width larger and smaller and make sure that any changes accommodates these different widths.

snaheth: Ahh. I've tried it out and it actually seems fine. Using height/2 places the text centered as we would like it:
screen shot 2016-11-10 at 2 17 02 pm

PR #450, opened 2017-01-12
Interval.isCoveredBy() fix

akmorrow13:
Fixed isCoveredBy() to pass for cases where the interval is not found in the first Interval queried

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/450)


armish:
Thanks for the fix, @akmorrow13!

PR #453, opened 2017-02-12
GA4GHDataSource only requests reads not in cache

akmorrow13:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/453)


akmorrow13:
The purpose of this commit is to avoid requesting regions that are already in the reads cache. An example where this is useful is if one requests ("chr17", 1000, 2000) and then moves to ("chr17", 800, 2000), this will request only  ("chr17", 800, 1000)  instead of  ("chr17", 800, 2000)

PR #454, opened 2017-02-26
Added Support for Building on Windows

chgibb:
Some minor changes to the build toolchain to support building of PileupJS on Windows. Confirmed works on Windows.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/454)

### DIFF
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,7 +1,7 @@
 #!/bin/bash
 # Build require-ale and minified assets for distribution.
 set -o errexit
-./scripts/make-mini-d3.sh # TODO: remove
+sh scripts/make-mini-d3.sh # TODO: remove
 
 # Transpile individual files. This is useful if another module,
 # e.g. cycledash, wants to require('pileup').
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 # Build require-ale and minified assets for distribution.
+(set -o igncr) 2>/dev/null && set -o igncr; # For Cygwin on Windows compatibility
 set -o errexit
 sh scripts/make-mini-d3.sh # TODO: remove
 
--- a/scripts/build.sh
+++ b/scripts/build.sh
@@ -7,11 +7,11 @@ sh scripts/make-mini-d3.sh # TODO: remove
 # e.g. cycledash, wants to require('pileup').
 # The dist/test files are required for code coverage
 mkdir -p dist/test/{data,source,viz}
-babel src --retain-lines --ignore src/lib --out-dir dist
+node_modules/.bin/babel src --retain-lines --ignore src/lib --out-dir dist
 cp -r src/lib dist/
 
 # Create dist/tests
-browserify \
+node_modules/.bin/browserify \
   -v \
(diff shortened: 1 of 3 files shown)


scripts/make-mini-d3.sh
```diff
@@ -1,12 +1,13 @@
 #!/bin/bash
+(set -o igncr) 2>/dev/null && set -o igncr; # For Cygwin on Windows compatibility
```
armish: can you explain what this does a bit more and maybe put a pointer to the comments? I am asking this because if we hit an issue because of this setting, we then can go back and check the details in the future.

scripts/build.sh
```diff
@@ -1,25 +1,26 @@
...
 set -o errexit
-./scripts/make-mini-d3.sh # TODO: remove
+sh scripts/make-mini-d3.sh # TODO: remove
 
 # Transpile individual files. This is useful if another module,
 # e.g. cycledash, wants to require('pileup').
 # The dist/test files are required for code coverage
 mkdir -p dist/test/{data,source,viz}
-babel src --retain-lines --ignore src/lib --out-dir dist
+node_modules/.bin/babel src --retain-lines --ignore src/lib --out-dir dist
```
chgibb: The path will not work for those who have these things installed globally, that is correct. 

On Windows it should not be run through NPM due to the ./ syntax used. That syntax is not supported by Cygwin. Instead of rewriting all the NPM scripts to play nice with Cygwin, I thought it was easier and less error prone to simply change build.sh.

As for the PATH adjustments, I'm not sure what you mean.

scripts/make-mini-d3.sh
chgibb: igncr is a Cygwin only setting which causes Cygwin bash to recognize \n\r as line endings as opposed to recognising the purely Linux \n as a line ending and then \r as an error.

This line attempts to set the option, suppresses any error messages. If it is not set successively then abort. If there is no error, then the option is set for real.

The snippet in question is from Eric Blake on the Cygwin mailing list from a few years back https://cygwin.com/ml/cygwin-announce/2010-08/msg00015.html this solution particularly seemed like the most painless to implement.

scripts/build.sh
armish: ah, gotcha! Didn't realize that the `npm run build` was failing due to the way `scripts` are defined in `package.json`.

I tried googling a solution for this, realized that this is a common issue, and it is NPM-related rather than pileup itself. Furthermore, NPM people officially do **not** support `cygwin` and recommend against using it: https://github.com/npm/npm#installing-on-cygwin

Although it would be great to support cygwin, I don't think we should make strong assumptions about people's setup and just do the things in the NPM way as much as possible. I didn't know about the `cygwin/npm` interaction was that problematic and I think it is not up to us to start adjusting things for a platform that is not supported. 

I really appreciate the PR and your contribution, but I don't think we should merge this into pileup.

armish:
Turns out NPM officially doesn't support cygwin: https://github.com/npm/npm#installing-on-cygwin hence we shouldn't try to work around cygwin-specific problems in this package. Closing the PR.

chgibb:
@armish fair enough. I appreciate the speedy replies and discussion!

PR #460, opened 2017-06-15
Added GA4GHAlignments to be accessible through raw json

akmorrow13:
This is primarily to allow pileup.js to be used in platforms other than the browser. For example, a user can load data into a pandas data frame in python, and pipe the dataframe using the  to_json() function into pileup. Since python notebook supports html and JS, one can immediately view their in the same environment. 

I think I will eventually add similar functionality for variants, features, etc. but thought this is a good starting point for discussion.

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/460)

### DIFF
--- a/src/main/json/GA4GHJson.js
+++ b/src/main/json/GA4GHJson.js
@@ -1,0 +1,50 @@
+/**
+ * A data source which implements generic JSON protocol.
+ * Currently only used to load alignments.
+ * @flow
+ */
+'use strict';
+
+import type {Alignment, AlignmentDataSource} from '../Alignment';
+
+import _ from 'underscore';
+import {Events} from 'backbone';
+
+import ContigInterval from '../ContigInterval';
+import GA4GHAlignment from '../GA4GHAlignment';
+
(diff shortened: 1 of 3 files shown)


armish:
Thanks, @akmorrow13! This is great — notebook integration has been something on our todo list since early days.

If you get it to work fully, would love to test it out and put an example on the repo.

akmorrow13:
Hi  @armish  ! Tagging @jpdna for interest. The reads visualization already works in a widget (with my personal pileup branch with these JSON changes). The widget repo can be found here: https://github.com/akmorrow13/genomicviz

This is more of a wrapper, with pileup.js being the main focus. However, we will be adding in some other exciting visualizations, such as heat maps and stuff. Any useful viz recommendations would be welcome.

I am working to add variants and features next. First though, I will be adding GA4GH formats for these types into pileup.js. PR is coming soon!

armish:
@akmorrow13 @jpdna - that is amazing news! I will keep an eye on the project in case you need beta testers :)

PR #464, opened 2017-07-04
Added Variants and Features GA4GH, with GA4GH example file

akmorrow13:
This PR contains the following:
- GA4GH Variants
- GA4GV Features
- Features Viz Track
- GA4GH Example File for Features and Variants

What's left for this PR:

- [x] Add a GA4GH Reads example on the GA4GH example page.
- [x] There is an issue with using sinon server and dataCanvas at the same time. Therefore, the tests for FeatureTrack will fail. We can either add in the Feature JSON and rewrite the tests on the JSON datasource, or add these tests in a separate PR once the JSON formats are done.
- [x] Test in python to verify GA4GH schemas

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/464)


src/main/Interval.js
```diff
@@ -39,6 +39,16 @@ class Interval {
     return this.contains(other.start) && this.contains(other.stop);
   }
 
+  // Expand range to begin and end on multiples of size
+  expand(size: number, zeroBased: boolean): Interval {
+    var minimum = zeroBased ? 0 : 1;
+    var roundDown = x => x - x % size;
```
armish: can you add a simple example to what this line would do as a comment? And maybe encapsulate the modulo operation in parentheses just for clarity?

This is just me thinking loud: the `expand` name seems a bit misleading here: I first thought that this was just expanding the region by the given size, but this might actually not expand one of the ends if either the `start` or `stop` are already multiples of a size - e.g.: `(10, 20) -> expand(5) -> (10, 20)`. Maybe using `round` in the name would be better but I leave it up to you since this was what it was named originally.

src/main/sources/BamDataSource.js
```diff
@@ -77,7 +70,7 @@ function createFromBamFile(remoteSource: BamFile): AlignmentDataSource {
         return Q.when();
       }
 
-      interval = expandRange(interval);
+      interval = interval.expand(BASE_PAIRS_PER_FETCH, ZERO_BASED);
```
armish: :+1:

src/main/sources/BigBedDataSource.js
```diff
@@ -22,10 +22,28 @@ export type Gene = {
...
+  on: (event: string, handler: Function) => void;
+  off: (event: string) => void;
+  trigger: (event: string, ...args:any) => void;
+}
+
 // Flow type for export.
 export type BigBedSource = {
   rangeChanged: (newRange: GenomeRange) => void;
-  getGenesInRange: (range: ContigInterval<string>) => Gene[];
+  getFeaturesInRange: (range: ContigInterval<string>) => Gene[];
```
armish: thanks for renaming this — `feature` would better speak to the bioinformaticians in this context.

src/main/sources/GA4GHAlignmentSource.js
```diff
@@ -52,6 +41,9 @@ function create(spec: GA4GHSpec): AlignmentDataSource {
   var coveredRanges: ContigInterval<string>[] = [];
 
   function addReadsFromResponse(response: Object) {
+    if (response.alignments === undefined) {
+      return;
```
armish: curious why/when this happens. I would have expected the `alignments` to be an empty array for the semantics of it, but is this how `GA4GH` server responds when no reads match the criteria?

src/main/viz/VariantTrack.js
```diff
@@ -126,7 +126,7 @@ class VariantTrack extends React.Component {
     if (variants && variants.length>0) {
       var data = [];
       for (var i=0;i<variants.length;i++) {
-        data.push({id: variants[i].id,vcfLine:variants[i].vcfLine});
+        data.push({id:variants[i].id,vcfLine:variants[i].vcfLine,ref:variants[i].ref,alt:variants[i].alt});
```
armish: sorry for the styling inconsistency here; if you had a chance, would you mind fixing the style here? OK if you don't get to it:

```javascript
// ...

for (var i=0; i < variants.length; i++) {
  data.push({
    id: variants[i].id,
    vcfLine: variants[i].vcfLine,
    ref: variants[i].ref,
    alt: variants[i].alt
  });
}

// ...
```

src/main/Interval.js
akmorrow13: Ok sounds good! I will rename it. Also, I am still not sure on the `zeroBased `parameter. Another option is changing `zeroBased` to an integer, and passing that in as the minimum. So instead you would have:

```
round(size: number, minimum: number): Interval {
    var roundDown = x => x - x % size;
    var newStart = Math.max(minimum, roundDown(this.start)),
        newStop = roundDown(this.stop + size - 1);
    return new Interval(newStart, newStop);
}
```
This is pretty minor, but it could be cleaner.

src/main/sources/GA4GHAlignmentSource.js
akmorrow13: This occurs when you query a region without data. For example, for variants, if you query the GA4GH server:
`curl --data '{"variantSetId": "WyIxa2dlbm9tZXMiLCJ2cyIsInBoYXNlMy1yZWxlYXNlIl0", "end": "10200", "pageSize": "0", "pageToken": "", "start": "10000", "callSetIds": ["WyIxa2dlbm9tZXMiLCJ2cyIsInBoYXNlMy1yZWxlYXNlIiwiSEcwMDA5NiJd"],  "referenceName": "10"}' --header 'Content-Type: application/json'     http://1kgenomes.ga4gh.org/variants/search`

It will return `{}`

Therefore, if trying to call `.variants `(or `.alignments` in the alignment case),  on the empty JSON object, the code will error.

src/main/viz/VariantTrack.js
akmorrow13: Styling nits are appreciated! Will fix!

akmorrow13:
@armish Thanks for all the comments! I decided that it would be easiest to just add the JSON stuff required to pass the FeatureTrack tests, so those should all be in! Let me know if there are any other comments we should address. 

One thing to note is the Flow issues: In VcfDataSource.js and BigBedDataSource.js I had to add an export flag to make Flow happy with some imports.  Although this is not the cleanest solution, it seems to be a bug with jsxhint that can only be fixed this way.

armish:
thank you, @akmorrow13! LGTM.

Issue #463, opened 2017-06-30
Layout options for viewing multiple tracks


Georgehe4:
Pileup does not support multiple tracks very well due to the flex layout. 
The tracks become squished and are hard to view. 
Are there any plans to address this use case? Having minimum height/pre-fixed track size/user-controllable track height might help resolve some of these issues.

armish:
Hey @Georgehe4 - hmm, that sound weird because unless there is no room at all for the tracks, flex-based sized elements shouldn't collapse. Since all those attributes are being defined in the CSS, you should be able to override them by setting minimums by yourself and not allow that to happen.

How many tracks are we talking about here by the way? And also would be great to have this somewhere like CodePen so I can inspect directly and don't make guesstimates. If that is hard, a screenshot would also help us get started.

Let me know (and sorry for the late reply)

armish:
addressed via #465.

PR #465, opened 2017-07-14
Set minimum size for track viewing

Georgehe4:
Follow up of #463

This change will set minimum height values for tracks to help prevent the use case where the window is too small to view anything, shown below
screen shot 2017-07-14 at 11 24 31 am

---
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/465)


armish:
thank you so much, @Georgehe4! Looks good.

Issue #477, opened 2018-06-22
Simple17.bb fails to display with `Cannot read property 'replace' of undefined`


denisemauldin:
When loading a simple bigBed file that doesn't contain exon arrays, pileup.js throws an error.

Cannot read property 'replace' of undefined
    at parseBedFeature (BigBedDataSource.js:51)

```
function parseBedFeature(f): Gene {
  var position = new ContigInterval(f.contig, f.start, f.stop),
      x = f.rest.split('\t'),
      // exons arrays sometimes have trailing commas
      exonLengths = x[7].replace(/,*$/, '').split(',').map(Number),
      exonStarts = x[8].replace(/,*$/, '').split(',').map(Number),
      exons = _.zip(exonStarts, exonLengths)
               .map(function([start, length]) {
                 return new Interval(f.start + start, f.start + start + length);
               });

  return {
    position,
    id: x[0],  // e.g. ENST00000359597
    strand: x[2],  // either + or -
    codingRegion: new Interval(Number(x[3]), Number(x[4])),
    geneId: x[9],
    name: x[10],
    exons
  };
}
```

This is the config I'm using:

```
{
                    viz: pileup.viz.genes(),
                    data: pileup.formats.bigBed({
                      url: 'http://localhost:3000/assets/data/simple17.bb'
                    }),
                    name: 'Simple17'
                },
```

How do I load a bigBed file that only has `chrom, chromStart, chromEnd, name, score, strand`?

denisemauldin:
I don't think it's the viz that's the problem.  I'm still getting the same error.  Is there a parser for regular bed files that isn't `pileup.formats.bigBed`?

akmorrow13:
I think this could be fixed by adding a new function in BigBedDataSource.js called getGenesInRange(), which returns Genes, and getFeaturesInRange, which returns Features. FeatureTrack in range then would return features instead of Genes. With these changes the following should work:

{
                    viz: pileup.viz.features(),
                    data: pileup.formats.bigBed({
                      url: 'http://localhost:3000/assets/data/simple17.bb'
                    }),
                    name: 'Simple17'
                },

Is there anyway you can share the first couple lines of your file?

akmorrow13:
https://github.com/hammerlab/pileup.js/pull/478/commits should allow this to work

denisemauldin:
@akmorrow13 I was almost done (fixing the testing) with the PR I was making, so I submitted it just in case it was useful so we could chat about it!

[...]

PR #478, opened 2018-06-26
fixed BigBedDataSource to work with FeatureTrack

akmorrow13:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/478)


akmorrow13:
Addresses https://github.com/hammerlab/pileup.js/issues/477

src/main/sources/BigBedDataSource.js
```diff
@@ -44,29 +45,44 @@ export type BigBedSource = {
...
-      exonStarts = x[8].replace(/,*$/, '').split(',').map(Number),
-      exons = _.zip(exonStarts, exonLengths)
-               .map(function([start, length]) {
-                 return new Interval(f.start + start, f.start + start + length);
-               });
+      x = f.rest.split('\t');
+
+  // if no id, generate randomly for unique storage
+  var id = x[0] ? x[0] : "feature_" + Math.random().toString(36).substr(2, 9); // e.g. ENST00000359597
+  var score = x[1] ? x[1] : 1000; // number from 0-1000
```
denisemauldin: Won't this color the feature randomly?

src/main/sources/BigBedDataSource.js
```diff
@@ -44,29 +45,44 @@ export type BigBedSource = {
...
+  var strand =  x[2] ? x[2] : '.'; // either +, - or .
+  var codingRegion = (x[3] && x[4]) ? new Interval(Number(x[3]), Number(x[4])) :new Interval(0, 0);
```
denisemauldin: Is it better to assume that the coding region matches the position?

src/main/sources/BigBedDataSource.js
```diff
@@ -44,29 +45,44 @@ export type BigBedSource = {
...
+  var geneId =  x[9] ? x[9] : "";
```
denisemauldin: I defaulted to having this be the `position.toString()` because otherwise nothing shows up as the label.

src/main/sources/BigBedDataSource.js
akmorrow13: Feature score will create an alpha that is in a range of 0-1000. If the score is not specified, alpha will be set to 1.

src/main/sources/BigBedDataSource.js
akmorrow13: ahh good call! let me update

src/main/sources/BigBedDataSource.js
akmorrow13: Yes! fixed.

PR #479, opened 2018-06-26
Update BED parsing

denisemauldin:
This is my version that I was working on before I saw yours @akmorrow13 !  Would you mind reviewing this anyway?  Do you think it would be useful for me to submit the tests as a separate PR?

### DIFF
--- a/src/main/sources/BigBedDataSource.js
+++ b/src/main/sources/BigBedDataSource.js
@@ -46,24 +46,31 @@ export type BigBedSource = {
 // The fields are described at http://genome.ucsc.edu/FAQ/FAQformat#format1
 function parseBedFeature(f): Gene {
   var position = new ContigInterval(f.contig, f.start, f.stop),
-      x = f.rest.split('\t'),
-      // exons arrays sometimes have trailing commas
-      exonLengths = x[7].replace(/,*$/, '').split(',').map(Number),
-      exonStarts = x[8].replace(/,*$/, '').split(',').map(Number),
-      exons = _.zip(exonStarts, exonLengths)
-               .map(function([start, length]) {
-                 return new Interval(f.start + start, f.start + start + length);
-               });
+      x = f.rest.split('\t'), exons = [];
 
-  return {
-    position,
-    id: x[0],  // e.g. ENST00000359597
-    strand: x[2],  // either + or -
-    codingRegion: new Interval(Number(x[3]), Number(x[4])),
-    geneId: x[9],
-    name: x[10],
-    exons
-  };
+	if (x.length > 0 && x[0] !== "") {
+		// exons arrays sometimes have trailing commas
+		var exonLengths = x[7].replace(/,*$/, '').split(',').map(Number),
+		exonStarts = x[8].replace(/,*$/, '').split(',').map(Number);
+		exons = _.zip(exonStarts, exonLengths)
+					.map(function([start, length]) {
+						return new Interval(f.start + start, f.start + start + length);
+					});
+
+    return {
+      position,
+      id: x[0],  // e.g. ENST00000359597
+      strand: x[2],  // either + or -
+      codingRegion: new Interval(Number(x[3]), Number(x[4])),
+      geneId: x[9],
+      name: x[10],
+      exons
+    };
+  } else {
+    return {
+      position
+    };
+  }
 }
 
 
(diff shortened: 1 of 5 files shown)


denisemauldin:
Addresses #477

denisemauldin:
@akmorrow13 Hmm... how do I get the Travis results to show up somewhere locally before I push?  I ran the tests, but it showed the same test failures locally as it did with the master branch...

akmorrow13:
@denisemauldin travis runs remotely so you can not run it locally. However, you can run all of the commands that travis runs, listed in the [.travis.yml file](https://github.com/hammerlab/pileup.js/blob/master/.travis.yml). It looks like `flow check` is failing for this PR.

src/main/sources/BigBedDataSource.js
```diff
@@ -46,24 +46,31 @@ export type BigBedSource = {
...
+		exons = _.zip(exonStarts, exonLengths)
+					.map(function([start, length]) {
+						return new Interval(f.start + start, f.start + start + length);
+					});
+
+    return {
+      position,
+      id: x[0],  // e.g. ENST00000359597
+      strand: x[2],  // either + or -
+      codingRegion: new Interval(Number(x[3]), Number(x[4])),
```
akmorrow13: This will still fail if you only have position information. x[0]-x[10] may be missing as well. I remember you mentioning that your bed files only had contig, start and end. Maybe I misunderstood.

src/main/viz/GeneTrack.js
```diff
@@ -57,7 +57,7 @@ function drawGeneName(ctx: CanvasRenderingContext2D,
                       textIntervals: Interval[]) {
   var p = gene.position,
       centerX = 0.5 * (clampedScale(1 + p.start()) + clampedScale(1 + p.stop()));
-  var name = gene.name || gene.id;
+  var name = gene.name || gene.id || gene.position.toString();
```
akmorrow13: We could actually put this in a separate PR, this seems useful.

src/main/viz/GeneTrack.js
```diff
@@ -144,23 +144,25 @@ class GeneTrack extends React.Component {
...
-      var introns = gene.position.interval.complementIntervals(gene.exons);
-      introns.forEach(range => {
-        drawArrow(ctx, clampedScale, range, geneLineY + 0.5, gene.strand);
-      });
-      ctx.strokeStyle = style.GENE_COMPLEMENT_COLOR;
-      ctx.lineWidth = 2;
-      gene.exons.forEach(range => {
-        drawArrow(ctx, clampedScale, range, geneLineY + 0.5, gene.strand);
-      });
+      if (gene.strand && gene.exons && gene.codingRegion) {
```
akmorrow13: I don't think we want to modify the GeneTrack code. If strand and exons and codingRegion are missing, FeatureTrack would be a better fit for the data.

denisemauldin:
Cool.  I've never done flow before.  I'll check it out 😄

src/main/viz/GeneTrack.js
denisemauldin: I agree that a FeatureTrack is a better fit for the data, but I don't know which data is expected for each visualization type.   I don't know when I'm supposed to use a FeatureTrack vs a CoverageTrack vs a GenomeTrack vs a GeneTrack when I'm a naive user with a BAM or BED file in hand.

I try to utilize defensive programming for user naivety and make sure something runs and doesn't throw errors if the user attempts to do something not intended by the code.  Wrapping this set of code in an if statement makes it so that the GeneTrack loads even when provided a simple BED file with 3 columns.

Without these changes, GeneTrack throws:
```
bedtools.js:21 Uncaught TypeError: Cannot read property 'map' of undefined
    at Object.splitCodingExons (bedtools.js:21)
    at GeneTrack.js:147
    at Array.forEach (<anonymous>)
    at GeneTrack.updateVisualization (GeneTrack.js:136)
    at GeneTrack.componentDidUpdate (GeneTrack.js:104)
    at CallbackQueue.notifyAll (CallbackQueue.js:65)
    at ReactReconcileTransaction.close (ReactReconcileTransaction.js:81)
    at ReactReconcileTransaction.closeAll (Transaction.js:202)
    at ReactReconcileTransaction.perform (Transaction.js:149)
    at ReactUpdatesFlushTransaction.perform (Transaction.js:136)
```

PR #482, opened 2018-06-27
Added bigbed test for simple bigBed file.

denisemauldin:
Updated GeneTrack to be defensive about bigBed files.
Added simple bigBed to data test.


denisemauldin:
@akmorrow13 I can remove the GeneTrack.js defensiveness in this commit if you'd like to go with a different route for providing new users a better experience.

PR #483, opened 2018-06-27
Added the ability to pass a function into pileup click handler

denisemauldin:
@akmorrow13 I can't tell how to get the click test to work.

`drawnObjectsWith(testDiv, '.pileup')` throws a failure even though if I comment out the other tests and remove the resets on testDiv and p, it displays fine in the browser.

There seems to be a timing issue because sometimes `var canvasList = testDiv.getElementsByTagName('canvas');` has two elements and sometimes it doesn't.  Maybe `hasPileupSelector` is not an appropriate wait for this test?

I don't know how to get what offset I should provide to the `ReactTestUtils.Simulate.click`.

Thanks for your help.

### DIFF
--- a/src/main/viz/PileupTrack.js
+++ b/src/main/viz/PileupTrack.js
@@ -475,8 +475,14 @@ class PileupTrack extends React.Component {
     renderPileup(trackingCtx, scale, range, null, false, vGroups);
     var vRead = _.find(trackingCtx.hits[0], hit => hit.read);
     var alert = window.alert || console.log;
+
     if (vRead) {
-      alert(vRead.read.debugString());
+      //user provided function for displaying popup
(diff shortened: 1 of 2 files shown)


akmorrow13:
@denisemauldin you can run the tests interactively by running the http server and going to test-runner (instructions here). 

```
npm run http-server
open http://localhost:8080/src/test/runner.html
```

I am not sure if you already tried this but it can help debug the ordering of events.

denisemauldin:
@akmorrow13 Yeah, that's what I've been doing - http://localhost:8080/src/test/runner.html?grep=FeatureTrack

If I comment out all of the other tests and the ` testDiv.innerHTML = '';` in the `afterEach` then it shows the pileup on the screen, but the `canvas` variable is still `undefined`.

PR #491, opened 2018-08-24
lint and tests passing with new flow version

akmorrow13:


akmorrow13:
The flow version was very out of date, breaking the travis build due to memory requirements that surpassed the travis limit. We need to update flow to 0.79.1, and all the new errors it has created. We also moved jsxhint to eslint to fix some resulting lint bugs.

PR #492, opened 2018-08-25
lint and tests passing with new flow version

akmorrow13:
This change is [Reviewable](https://reviewable.io/reviews/hammerlab/pileup.js/492)


akmorrow13:
This PR addresses the travis break from flow. The previous flow version (^0.21.0) exceeded the travis memory limits. We updated the flow version to ^0.79.1. To update flow, we needed to update a bunch of checks in the code. Many of these checks were checking for null values, correct function headers, etc. We also had to update the travis node.js version from 5.1 to 6.14 for lint to run correctly.

PR #509, opened 2019-03-06
Fix mobile display cutting off reference genome letters

chfzhang:
On devices with retina displays (and simulated mobile displays in browser debug), letters for the reference genome would appear cutoff when zooming in. This fixes it by resizing the font for the letters based on screen size.


chfzhang:
Whoops, need to clean up commit history.

PR #510, opened 2019-03-06
Fix mobile display cutting off reference genome letters

chfzhang:
On devices with retina displays (and simulated mobile displays in browser debug), letters for the reference genome would appear cutoff when zooming in. This fixes it by resizing the font for the letters based on screen size.

### DIFF
--- a/src/main/viz/GenomeTrack.js
+++ b/src/main/viz/GenomeTrack.js
@@ -35,10 +35,11 @@ function renderGenome(ctx: DataCanvasRenderingContext2D,
   if (mode != DisplayMode.HIDDEN) {
     ctx.textAlign = 'center';
     if (mode == DisplayMode.LOOSE) {
-      ctx.font = style.LOOSE_TEXT_STYLE;
+      //ctx.font = style.LOOSE_TEXT_STYLE;
     } else if (mode == DisplayMode.TIGHT) {
-      ctx.font = style.TIGHT_TEXT_STYLE;
+      //ctx.font = style.TIGHT_TEXT_STYLE;
     }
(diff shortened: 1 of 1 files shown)


src/main/viz/GenomeTrack.js
```diff
@@ -35,10 +35,11 @@ function renderGenome(ctx: DataCanvasRenderingContext2D,
   if (mode != DisplayMode.HIDDEN) {
     ctx.textAlign = 'center';
     if (mode == DisplayMode.LOOSE) {
-      ctx.font = style.LOOSE_TEXT_STYLE;
+      //ctx.font = style.LOOSE_TEXT_STYLE;
     } else if (mode == DisplayMode.TIGHT) {
-      ctx.font = style.TIGHT_TEXT_STYLE;
+      //ctx.font = style.TIGHT_TEXT_STYLE;
     }
+    ctx.font = String(Math.min(pxPerLetter, 12)) + "px" + ` 'Helvetica Neue', Helvetica, Arial, sans-serif`;
```
akmorrow13: pxPerLetter is for x axis size adjustment, not height adjustment. See https://github.com/hammerlab/pileup.js/blob/master/src/main/viz/GenomeTrack.js#L65, where they use pxPerLetter to size the rectangle widths.

This will not work for all cases (see screenshot below)

screen shot 2019-03-06 at 10 41 44 am

You will need to re-adjust where the letters are printed : 

```
 ctx.textBaseline = "middle";
 ctx.fillText(letter, scale(1 + 0.5 + pos), adjustedHeight/2); 
```

where adjustedHeight is calculated from a scaling factor based on the display being used.

akmorrow13:
Thanks @chfzhang ! Unfortunately, this does not always fix the problem. I left a review for some suggestions that may help fix all cases.

akmorrow13:
Closed in favor of https://github.com/hammerlab/pileup.js/pull/511

PR #511, opened 2019-03-17
fixed hidden genome track letters

akmorrow13:


akmorrow13:
GenomeTrack letters were cutoff because font size was not rescaled to the window pixel ratio. This PR resizes text for GenomeTrack based on the device.

PR #513, opened 2019-04-10
features height is not 0

geenakim:


examples/data.js
```diff
@@ -6,6 +6,10 @@ var bamSource = pileup.formats.bam({
   indexUrl: '/test-data/synth3.normal.17.7500000-7515000.bam.bai'
 });
 
+var featureSource = pileup.formats.bigBed({
```
akmorrow13: Can you remove these so that the example file is not changed?

examples/data.js
```diff
@@ -53,6 +57,12 @@ var sources = [
     cssClass: 'normal',
     name: 'Coverage'
   },
+  {
```
akmorrow13: See comment above

src/main/viz/FeatureTrack.js
```diff
@@ -45,7 +45,7 @@ class FeatureTiledCanvas extends TiledCanvas {
   }
 
   heightForRef(ref: string): number {
-    return this.cache.pileupHeightForRef(ref) *
+    return Math.max(this.cache.pileupHeightForRef(ref),1) *
```
akmorrow13: Why is Math.max required twice?

style/pileup.css
```diff
@@ -191,6 +191,8 @@
 
 /* features track */
 .pileup-root > .features {
+  /*flex: 1;*/  /* stretch to fill remaining space */ /* fills too much space */
```
akmorrow13: Remove comment. Can you leave a comment on why `flex-basis: auto` is working?

style/pileup.css
```diff
@@ -191,6 +191,8 @@
 
 /* features track */
 .pileup-root > .features {
+  /*flex: 1;*/
```
akmorrow13: Can you remove this if we are no longer using it?

akmorrow13:
Thanks @geenakim!

Issue #494, opened 2018-08-30
TypeError: vRead.read.debugString is not a function


akmorrow13:
errors in PileupTrack at https://github.com/hammerlab/pileup.js/blob/master/src/main/viz/PileupTrack.js#L469

PR #516, opened 2019-04-22
debugString function added

abhim00:


scripts/test.sh
```diff
@@ -18,4 +18,4 @@ trap finish EXIT
 sleep 1
 
 # Start the tests
-mocha-phantomjs http://localhost:8081/src/test/runner.html "$@"
+mocha-phantomjs -p /usr/local/bin/phantomjs http://localhost:8081/src/test/runner.html "$@"
```
akmorrow13: this file should not be modified.

src/test/GA4GHAlignment-test.js
```diff
@@ -36,6 +36,8 @@ describe('GA4GHAlignment', function() {
       pos: 10007,
       strand: '-'
     });
+    expect(a.debugString().length).to.be.above(0);
+    console.log("GA4GHAlignment : " + a.debugString());
```
akmorrow13: remove console statement

src/test/data/SamRead-test.js
```diff
@@ -44,6 +44,8 @@ describe('SamRead', function() {
         pos: 79,
         strand: '-'
       });
+      expect(read.debugString().length).to.be.above(0);
+      console.log("SamRead : " + read.debugString());
```
akmorrow13: remove console statement

akmorrow13:
Fixes https://github.com/hammerlab/pileup.js/issues/494

src/main/GA4GHAlignment.js
```diff
@@ -118,5 +128,8 @@ class GA4GHAlignment /* implements Alignment */ {
     return alignment.fragmentName + ':' + alignment.readNumber;
   }
 }
+function makeCigarString(cigarOps: Array<CigarOp>) {
```
akmorrow13: It looks like this function is copied from [SamRead.js](https://github.com/hammerlab/pileup.js/blob/master/src/main/data/SamRead.js#L253). Can you move this function to Alignment.js from both SamRead and GA4GHAlignment and reference the function from Alignment.js?

abhim00: The function is actually not directly copied from SamRead.js. The function signatures are different in GA4GHAlignment.js and SamRead.js

akmorrow13: It would be great if we could actually change the SamRead function to match yours, considering it enforces types. We can still create a function in Alignment.js and modify the usage in SamRead.js to call the function in Alignment.js

abhim00: I made the edits, I think it should work.

akmorrow13:
Thanks @abhim00 !

PR #519, opened 2019-05-02
after changing pythonjs to mocha-headless-chrome grep on unit tests stopped working

piotr-gawron:
After changing pythonjs to mocha-headless-chrome there was no way to run single test from the console. 

For instance this call executed all unit tests, but shouldn't.
```
npm run test -- --grep=pileuputils
```
This PR restores grep functionality when running unit tests.

### DIFF
--- a/scripts/test.sh
+++ b/scripts/test.sh
@@ -17,5 +17,19 @@ trap finish EXIT
 # possible racing conditions
 sleep 1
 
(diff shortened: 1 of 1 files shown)


akmorrow13:
Thanks for the catch, @piotr-gawron! I tested it out and it works great. Thanks for the contribution!

PR #520, opened 2019-05-02
Big bed file with more than 256 contigs is handled properly.

piotr-gawron:
Current implementation of BigBed file doesn't handle properly files with more than 256 contigs. This is due to the assumption that contigs are a flat structure (https://github.com/hammerlab/pileup.js/blob/e99848dfd6cede20dbc3e13e411cfeee977e32eb/src/main/data/BigBed.js#L35)

This PR resolve the issue by parsing contig tree structure recursively if necessary.

I hope it will meet your coding standards. I couldn't find a way to enforce jbinary to parse dynamic recursive structures (but it might be doable). Therefore, when the situation arises I manually call jbinary to parse nodes recursively.


piotr-gawron:
Build failed due to https://github.com/indexzero/http-server/issues/521 and https://github.com/jfhbrook/node-ecstatic/issues/255

thornjad:
jfhbrook/node-ecstatic#256

akmorrow13:
Thanks for the PR @piotr-gawron! Looks like you all are working on backward compatability right now with the http-server. Once that gets in I can re-trigger the tests and take another look at this.

piotr-gawron:
Dear @akmorrow13,
I upgraded problematic dependency and re-triggered  CI.

Now travis is failing, but I'm not sure I understand why - is the `flow check` out of date?. 
The unit test is written properly - for simplicity you can return a promise instead of calling done callback (https://mochajs.org/#working-with-promises). Moreover, `_.keys` is a valid function in underscore (otherwise the test would fail).

src/test/data/BigBed-test.js
```diff
@@ -147,6 +148,13 @@ describe('BigBed', function() {
       });
   });
 
+  it('file with 257 contigs', function () {
```
akmorrow13: Tests need a return type to pass flow. To fix this, change `function ()` to `function (): any`

src/test/data/BigBed-test.js
```diff
@@ -147,6 +148,13 @@ describe('BigBed', function() {
...
+    var bigBed = new BigBed('/test-data/257-contigs.bb');
+    return bigBed.contigMap.then(contigs => {
+      expect(_.keys(contigs).length).to.equal(257);
```
akmorrow13: Flow needs this function to be declared in lib/underscore.js. You can add it there, and use [the values declaration as an example](https://github.com/hammerlab/pileup.js/blob/master/lib/underscore.js#L47)

src/test/components-test.js
```diff
@@ -9,52 +9,54 @@ import dataCanvas from 'data-canvas';
 import pileup from '../main/pileup';
 import ContigInterval from '../main/ContigInterval';
 import {waitFor} from './async';
-    
+
 describe('pileup', function() {
-  var tracks = [
```
akmorrow13: Does this file need to be modified? If not, can you revert it?

akmorrow13:
Thanks for the update @piotr-gawron ! I added some comments in the review for how you can fix the flow errors.

src/test/components-test.js
piotr-gawron: This is not a pure variable - it triggers downloading files and creating complex data structure that are used only in one unit test. I transformed it to function, but I would move it completely to unit test using it.
When I was developing/testing this initializer was poluting tests that should stay unaffected.

piotr-gawron:
@akmorrow13, I fixed flow issues. Let me know if you have more comments.

src/main/data/BigBed.js
```diff
@@ -30,21 +30,59 @@ function parseCirTree(buffer) {
...
+  //use a copy of the global definition and replace variable with constant (this variable would be unavailable here)
+  var typeCopy = _.extend({}, bbi.TYPE_SET);
+  typeCopy['BPlusTreeNode']['contents'][1][2]['key'][1] = keySize;
+  typeCopy['BPlusTreeNode']['contents'][1][3]['key'][1] = keySize;
+  return new jBinary(buffer, typeCopy).read('BPlusTreeNode');
 }
 
 // Generate the reverse map from contig ID --> contig name.
-function reverseContigMap(contigMap: {[key:string]: number}): Array<string> {
+function reverseContigMap(contigMap: { [key: string]: number }): Array<string> {
```
akmorrow13: It looks like your editor modifier a lot of the spacing/indentation in this file. Would you be able to revert the lines you did not intend to modify?

piotr-gawron: Sorry for that.
I reverted the indentation

akmorrow13:
Thanks @piotr-gawron!

PR #549, opened 2021-05-06
problem with chromosome mapping in vcf

piotr-gawron:
This PR address small issue with mapping chromosomes ids to numbers. When vcf file contained chromosomes like `chr20` the id was mapped to `0` instead of `20`.

There is also a unit test showing the issue.

### DIFF
--- a/src/main/data/vcf.js
+++ b/src/main/data/vcf.js
@@ -152,7 +152,7 @@ class ImmediateVcfFile {
 
     var contigMap = {};
     contigs.forEach(contig => {
-      if (contig.slice(0, 3) === 'chr') {
+      if (contig.slice(0, 3) == 'chr') {
         contigMap[contig.slice(3)] = contig;
       } else {
         contigMap['chr' + contig] = contig;
--- a/src/main/data/vcf.js
+++ b/src/main/data/vcf.js
@@ -152,8 +152,8 @@ class ImmediateVcfFile {
 
     var contigMap = {};
     contigs.forEach(contig => {
-      if (contig.slice(0, 3) == 'chr') {
-        contigMap[contig.slice(4)] = contig;
+      if (contig.slice(0, 3) === 'chr') {
+        contigMap[contig.slice(3)] = contig;
       } else {
         contigMap['chr' + contig] = contig;
       }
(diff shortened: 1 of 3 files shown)


src/main/data/vcf.js
```diff
@@ -152,8 +152,8 @@ class ImmediateVcfFile {
 
     var contigMap = {};
     contigs.forEach(contig => {
-      if (contig.slice(0, 3) == 'chr') {
-        contigMap[contig.slice(4)] = contig;
+      if (contig.slice(0, 3) === 'chr') {
```
akmorrow13, line 155: Why did you change to strict equality here?

akmorrow13 (changes requested):
Great catch @piotr-gawron! One small question, other then that looks great! Thank you for your contribution.

src/main/data/vcf.js
piotr-gawron: I did it probably out of habit. I never use `==` in JavaScript. There are too many pitfalls. My IDE also complains about `==` most of the time.

Probably in this case it does not matter so please revert. I did not want to interfere with your code style.

---
Another thing. I'm currently working on two features for pileup:
* support for tabix index in vcf file. So you would be able to visualize huge vcf files without network performance issue. There is a javascript library that deals with tabix indexes (https://github.com/GMOD/tabix-js). I'm just in a process of making it work in browserify (there are some dependencies that must be fixed).
* support for custom track visualization - in a project I'm working on there is need for visualization of  project-related data. I can fork from pileup and modify the code (in fact I already have it working). However, I was thinking about making it more generic, so user would be able to plug in to canvas drawing of his own track.

I hope the functionalities could be introduced in the main pileup.

akmorrow13: This is great @piotr-gawron! Please let me know if there is anything I can do to help with these features.

I am curious, what custom track visualizations have you built? Do you think any of them would be useful for bringing into pileup.js?

akmorrow13 (approved):

src/main/data/vcf.js
akmorrow13: Great! Very excited to see this @piotr-gawron!

